PR #1, opened 2021-11-29
Create plugin structure

nils-werner-sonarsource:


sonar-text-plugin/src/main/java/org/sonar/plugins/text/checks/AbstractInputStreamCheck.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Abstract Check to commonly catch possible IOExceptions for InputFiles
+ */
+public abstract class AbstractInputStreamCheck implements CommonCheck {
+
+  @Override
+  public void initialize(InitContext init) {
+    init.register((ctx, inputFile) -> {
+      try {
+        analyzeStream(ctx, inputFile.inputStream());
```
karim-ouerghemmi-sonarsource, line 37: The doc of `InputFile#inputStream()` says that the caller should close the stream ([link](https://github.com/SonarSource/sonarqube/blob/d91ea92fe22cf1b6b52a8efba122daf8e7e58282/sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/InputFile.java#L121-L128)). Let's do this here. We can change this into a try-with-resources (`try (InputStream = inputFile.inputStream()) {...}`)

karim-ouerghemmi-sonarsource (approved):
LGTM. See comments.

PR #10, opened 2021-11-30
Add BIDI rule check

karim-ouerghemmi-sonarsource:


sonar-text-plugin/src/main/java/org/sonar/plugins/text/checks/BIDICharacterCheck.java
```diff
@@ -0,0 +1,65 @@
...
+      checkLine(ctx, scanner.nextLine(), lineNumber);
+    }
+  }
+
+  private static void checkLine(CheckContext ctx, String lineContent, int lineNumber) {
+    for (Character bidiChar : BIDI_CHARS) {
+      if (lineContent.indexOf(bidiChar) >= 0) {
+        ctx.reportLineIssue(lineNumber, MESSAGE);
+      }
+    }
```
nils-werner-sonarsource, line 63: Wouldn't it be faster to have a regex with a character class looking for a BIDI char in the line?
```suggestion
    if (BIDI_CHARS_PATTER.matcher(lineContent).matches()) {
      ctx.reportLineIssue(lineNumber, MESSAGE);
    }
```
with
```
  private static final Pattern BIDI_CHARS_PATTER;
  static {
    List<String> chars = BIDI_CHARS.stream().map(String::valueOf).collect(Collectors.toList());
    BIDI_CHARS_PATTER = Pattern.compile(String.format("[%s]", String.join("|", chars)));
  }
```
WDYT?

nils-werner-sonarsource: If not, I would stop the loop on a match
```suggestion
    for (Character bidiChar : BIDI_CHARS) {
      if (lineContent.indexOf(bidiChar) >= 0) {
        ctx.reportLineIssue(lineNumber, MESSAGE);
        return;
      }
    }
```

nils-werner-sonarsource (approved):
LGTM! See comment

sonar-text-plugin/src/main/java/org/sonar/plugins/text/checks/BIDICharacterCheck.java
karim-ouerghemmi-sonarsource: I did try it out with regex as you proposed now, and the characters were not detected in the test.
Not sure why, but before investigating I did measure if it makes a difference on big files (I took tcpdf with its around 26k lines), and both approaches were more or less the same (regex a bit slower), but we are talking in 10th of milliseconds here (e.g., 98ms vs 108ms), so I don't think it's worth investigating now.

PR #11, opened 2021-11-30
Simplify CheckContext handling

nils-werner-sonarsource:


sonar-text-plugin/src/main/java/org/sonar/plugins/text/checks/BIDICharacterCheck.java
```diff
@@ -46,8 +48,17 @@
   );
 
   @Override
-  void analyzeStream(CheckContext ctx, InputFile inputFile, InputStream stream) {
-    Scanner scanner = new Scanner(stream, inputFile.charset().name());
+  public void analyze(InputFile inputFile) {
+    try (InputStream stream = inputFile.inputStream()) {
+      analyzeStream(stream, inputFile.charset());
+    } catch (IOException e) {
+      e.printStackTrace();
```
karim-ouerghemmi-sonarsource, line 55: Besides to what is already said in the security hotspot raised at this location, I think this exception should ideally raise up to the sensors `analyseFiles()` and be logged as an analysis warning there.

karim-ouerghemmi-sonarsource (approved):
LGTM. See comment about security hotspot.

PR #19, opened 2021-12-03
Fix promote task

nils-werner-sonarsource:
Fix a copy/paste mistake in the promote task and restructure the way we do our QA. Now we hide the `its` module to being part of the main build as well as the WhiteSource scan.


karim-ouerghemmi-sonarsource (approved):
LGTM!

PR #22, opened 2021-12-06
Adapt BIDI rule to reduce FPs

karim-ouerghemmi-sonarsource:


nils-werner-sonarsource (approved):
Very well solved. Extremely good PR with comments in places where they are really helpful.

PR #29, opened 2022-01-18
Setup release workflow

nils-werner-sonarsource:


.github/workflows/release.yml
```diff
@@ -0,0 +1,79 @@
...
+          CIRRUS_TOKEN: ${{ secrets.CIRRUS_TOKEN }}
+          PATH_PREFIX: ${{ secrets.BINARIES_PATH_PREFIX }}
+          GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
+          RELEASE_SSH_USER: ${{ secrets.RELEASE_SSH_USER }}
+          RELEASE_SSH_KEY: ${{ secrets.RELEASE_SSH_KEY }}
+          SLACK_API_TOKEN: ${{secrets.SLACK_API_TOKEN }}
+        uses: SonarSource/gh-action_release/main@master
+        with:
+          publish_to_binaries: true
+          attach_artifacts_to_github_release: true
```
tomverin, line 28: I've just updated the documentation, this option is not available anymore

.github/workflows/release.yml
```diff
@@ -0,0 +1,79 @@
...
+          distribute: true
+          distribute_target: false
+          run_rules_cov: false
```
tomverin, line 31: Same here, you should remove this option

.github/workflows/release.yml
```diff
@@ -0,0 +1,79 @@
...
+          slack_channel: team-lang-phpy
```
tomverin, line 32: are you sure about the y at the end?

tomverin (changes requested):

.github/workflows/release.yml
nils-werner-sonarsource: Yeah. We merged PHP and Python bubble and call it PHPy.

tomverin (approved):

PR #63, opened 2023-01-02
SONARTEXT-1 Import the 7 SonarSecrets rules into SonarText

alban-auzeill:


sonarpedia-secrets.json
```diff
@@ -0,0 +1,11 @@
+{
+  "rules-metadata-path": "./sonar-text-plugin/src/main/resources/org/sonar/l10n/secrets/rules/secrets",
```
Wohops, line 2: I think it would be easier to do the same as what was done on sonarjs with CSS: Their is only a small folder with the sonarpedia file, and they keep one at the root. See https://github.com/SonarSource/SonarJS/tree/master/css-sonarpedia

sonar-text-plugin/src/main/java/org/sonar/plugins/common/Check.java
```diff
@@ -0,0 +1,45 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.common;
+
+import java.io.IOException;
+import org.sonar.api.rule.RuleKey;
+import org.sonar.api.utils.AnnotationUtils;
+import org.sonar.check.Rule;
+
+public interface Check {
```
Wohops, line 27: Not sure that this needs to be an interface. Couldn't it be an abstract class that would cache the ruleKey and only compute it once?

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SecretsMatcher.java
```diff
@@ -0,0 +1,26 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.secrets.api;
+
+import java.util.List;
+
+public interface SecretsMatcher {
+  List<Match> findIn(String content);
```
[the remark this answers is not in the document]
Wohops: Was already there in sonar-secrets. Let's keep it like this.

its/ruling/src/test/java/org/sonarsource/text/TextRulingTest.java
```diff
@@ -55,15 +55,18 @@ public static void prepare_quality_profile() {
     ProfileGenerator.RulesConfiguration parameters = new ProfileGenerator.RulesConfiguration();
 
     String serverUrl = ORCHESTRATOR.getServer().getUrl();
-    File profileFile = ProfileGenerator.generateProfile(serverUrl, "text", "text", parameters, Collections.emptySet());
-    ORCHESTRATOR.getServer().restoreProfile(FileLocation.of(profileFile));
+    File textProfileFile = ProfileGenerator.generateProfile(serverUrl, "text", "text", parameters, Collections.emptySet());
+    ORCHESTRATOR.getServer().restoreProfile(FileLocation.of(textProfileFile));
+    File secretsProfileFile = ProfileGenerator.generateProfile(serverUrl, "secrets", "secrets", parameters, Collections.emptySet());
+    ORCHESTRATOR.getServer().restoreProfile(FileLocation.of(secretsProfileFile));
     ORCHESTRATOR.getServer().restoreProfile(FileLocation.of("src/test/resources/no_rules_php.xml"));
```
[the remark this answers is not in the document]
alban-auzeill: it's possible to remove sonar-php from its/ruling but not from its/plugin because of a test with NOSONAR

its/plugin/src/test/java/org/sonarsource/text/Tests.java
```diff
@@ -48,6 +48,8 @@ public class Tests {
       .useDefaultAdminCredentialsForBuilds(true)
       .setSonarVersion(System.getProperty(SQ_VERSION_PROPERTY, DEFAULT_SQ_VERSION))
       .addPlugin(PLUGIN_LOCATION)
+      // sonar-php-plugin is used to bring the NOSONAR filter to php files.
```
Wohops, line 51: What about creating a ticket for this so it's part of the backlog?

Wohops (approved):
LGTM! Nice work! Let's make peach explode!

PR #72, opened 2023-01-16
SONARTEXT-1 Fix source encoding WARN by converting BinaryFilePredicate as a org.sonar.api.batch.fs.FilePredicate

alban-auzeill:


johann-beleites-sonarsource (approved):
LGTM!

PR #81, opened 2023-01-30
Scaffold sonar-text-dotnet pipeline

mary-georgiou-sonarsource:
1. Update UTs to successfully run in the new sonar-text context.
2. Basic setup for the `sonar-dotnet-text` pipeline including.
    -  Create a windows VM in cirrus
    - Restore and build sonar-dotnet-text
    - run UTs

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -22,6 +22,7 @@ env:
 
 win_vm_definition: &WINDOWS_VM_DEFINITION
   ec2_instance:
+    experimental: true # see https://github.com/cirruslabs/cirrus-ci-docs/issues/1051
     image: base-windows-dotnet-v*
     region: eu-central-1
     subnet_id: ${CIRRUS_AWS_SUBNET}
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,115 +20,137 @@ env:
   # Allows to run builds for the 50 last commits in a branch:
   CIRRUS_CLONE_DEPTH: 1
 
-linux_image_template: &LINUX_IMAGE
-  image: ${CIRRUS_AWS_ACCOUNT}.dkr.ecr.eu-central-1.amazonaws.com/base:j11-latest
-  cluster_name: ${CIRRUS_CLUSTER_NAME}
-  region: eu-central-1
-  namespace: default
-  use_in_memory_disk: true
+win_vm_definition: &WINDOWS_VM_DEFINITION
+  ec2_instance:
+    experimental: true # see https://github.com/cirruslabs/cirrus-ci-docs/issues/1051
+    image: base-windows-dotnet-v*
+    region: eu-central-1
+    subnet_id: ${CIRRUS_AWS_SUBNET}
+    type: t2.medium # see https://aws.amazon.com/ec2/instance-types/
+    platform: windows
 
-linux_1_cpu_1G_template: &LINUX_1_CPU_1G
-  eks_container:
-    <<: *LINUX_IMAGE
-    cpu: 1
-    memory: 1G
+#linux_image_template: &LINUX_IMAGE
+#  image: ${CIRRUS_AWS_ACCOUNT}.dkr.ecr.eu-central-1.amazonaws.com/base:j11-latest
+#  cluster_name: ${CIRRUS_CLUSTER_NAME}
+#  region: eu-central-1
+#  namespace: default
+#  use_in_memory_disk: true
+#
+#linux_1_cpu_1G_template: &LINUX_1_CPU_1G
+#  eks_container:
+#    <<: *LINUX_IMAGE
+#    cpu: 1
+#    memory: 1G
+#
+#linux_2_cpu_4G_java_17_template: &LINUX_2_CPU_4G_JAVA_17
+#  eks_container:
+#    <<: *LINUX_IMAGE
+#    image: ${CIRRUS_AWS_ACCOUNT}.dkr.ecr.eu-central-1.amazonaws.com/base:j17-latest
+#    cpu: 2
+#    memory: 4G
+#
+#only_sonarsource_qa: &ONLY_IF_SONARSOURCE_QA
+#  only_if: $CIRRUS_USER_COLLABORATOR == 'true' && $CIRRUS_TAG == "" && ($CIRRUS_PR != "" || $CIRRUS_BRANCH == "master" || $CIRRUS_BRANCH =~ "branch-.*" || $CIRRUS_BRANCH =~ "dogfood-on-.*")
+#
+#qa_task_filter_template: &QA_TASK_FILTER
+#  depends_on:
+#    - build
+#  <<: *ONLY_IF_SONARSOURCE_QA
+#
+#build_task:
+#  <<: *ONLY_IF_SONARSOURCE_QA
+#  <<: *LINUX_1_CPU_1G
+#  env:
+#    SIGN_KEY: VAULT[development/kv/data/sign data.key]
+#    PGP_PASSPHRASE: VAULT[development/kv/data/sign data.passphrase]
+#    # analysis on next
+#    SONAR_TOKEN: VAULT[development/kv/data/next data.token]
+#    SONAR_HOST_URL: https://next.sonarqube.com/sonarqube
+#    #allow deployment of pull request artifacts to repox
+#    DEPLOY_PULL_REQUEST: true
+#  maven_cache:
+#    folder: ${CIRRUS_WORKING_DIR}/.m2/repository
+#  build_script:
+#    - source cirrus-env BUILD
+#    - regular_mvn_build_deploy_analyze
+#  cleanup_before_cache_script: cleanup_maven_repository
+#
+#ws_scan_task:
+#  depends_on:
+#    - build
+#  <<: *LINUX_1_CPU_1G
+#  # run only on master and long-term branches
+#  only_if: $CIRRUS_USER_COLLABORATOR == 'true' && $CIRRUS_TAG == "" && ($CIRRUS_BRANCH == "master" || $CIRRUS_BRANCH =~ "branch-.*")
+#  env:
+#    WS_APIKEY: VAULT[development/kv/data/mend data.apikey]
+#  maven_cache:
+#    folder: ${CIRRUS_WORKING_DIR}/.m2/repository
+#  whitesource_script:
+#    - source cirrus-env QA
+#    - source set_maven_build_version $BUILD_NUMBER
+#    - mvn clean install -DskipTests
+#    - source ws_scan.sh
+#  allow_failures: "true"
+#  always:
+#    ws_artifacts:
+#      path: "whitesource/**/*"
+#
+#qa_plugin_task:
+#  <<: *QA_TASK_FILTER
+#  <<: *LINUX_2_CPU_4G_JAVA_17
+#  env:
+#    matrix:
+#      - SQ_VERSION: LATEST_RELEASE
+#      - SQ_VERSION: DEV
+#  maven_cache:
+#    folder: ${CIRRUS_WORKING_DIR}/.m2/repository
+#  plugin_script:
+#    - source cirrus-env QA
+#    - source set_maven_build_version $BUILD_NUMBER
+#    - cd its/plugin
+#    - mvn verify -Drevision=${PROJECT_VERSION} -Dsonar.runtimeVersion=${SQ_VERSION} -B -e -V
+#  cleanup_before_cache_script: cleanup_maven_repository
+#
+#qa_ruling_task:
+#  <<: *QA_TASK_FILTER
+#  <<: *LINUX_2_CPU_4G_JAVA_17
+#  maven_cache:
+#    folder: ${CIRRUS_WORKING_DIR}/.m2/repository
+#  ruling_script:
+#    - source cirrus-env QA
+#    - source set_maven_build_version $BUILD_NUMBER
+#    - cd its/ruling
+#    - mvn verify -Dsonar.runtimeVersion=LATEST_RELEASE -B -e -V
+#  cleanup_before_cache_script: cleanup_maven_repository
+#
+#promote_task:
+#  depends_on:
+#    - build
+#    - qa_plugin
+#    - ws_scan
+#  <<: *ONLY_IF_SONARSOURCE_QA
+#  <<: *LINUX_1_CPU_1G
+#  env:
+#    #promotion cloud function
+#    GCF_ACCESS_TOKEN: VAULT[development/kv/data/promote data.token]
+#    PROMOTE_URL: VAULT[development/kv/data/promote data.url]
+#    GITHUB_TOKEN: VAULT[development/github/token/${CIRRUS_REPO_OWNER}-${CIRRUS_REPO_NAME}-promotion token]
+#    #artifacts that will have downloadable links in burgr
+#    ARTIFACTS: org.sonarsource.text:sonar-text-plugin:jar
+#  maven_cache:
+#    folder: ${CIRRUS_WORKING_DIR}/.m2/repository
+#  script: cirrus_promote_maven
+#  cleanup_before_cache_script: cleanup_maven_repository
 
-linux_2_cpu_4G_java_17_template: &LINUX_2_CPU_4G_JAVA_17
-  eks_container:
-    <<: *LINUX_IMAGE
-    image: ${CIRRUS_AWS_ACCOUNT}.dkr.ecr.eu-central-1.amazonaws.com/base:j17-latest
-    cpu: 2
-    memory: 4G
-
-only_sonarsource_qa: &ONLY_IF_SONARSOURCE_QA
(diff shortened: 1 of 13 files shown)


sonar-text-dotnet/pipeline_scripts/build_and_restore_solution.ps1
```diff
@@ -0,0 +1,26 @@
...
+
+$signArtifacts=$false
+if(($env:CIRRUS_BRANCH -eq "master" ) -or ($env:CIRRUS_BRANCH.StartsWith("branch-")))
+{
+    $signArtifacts=$true
+    Write-Host "signArtifacts variable was set to {$signArtifacts}".
+}
+
+Write-Host Building $env:SOLUTION_DIR
+dotnet build `
```
mary-georgiou-sonarsource, line 16: FYI not all build parameters are here, there's a task to add missing ones https://trello.com/c/EHSnIq17.

.cirrus.yml
```diff
@@ -132,3 +141,18 @@ promote_task:
...
+  depends_on:
+    - promote
+  <<: *ONLY_IF_SONARSOURCE_QA
+  <<: *WINDOWS_VM_DEFINITION
+  env:
+    PROJECT_DIR: ${CIRRUS_WORKING_DIR}\\sonar-text-dotnet
+    SOLUTION_DIR: ${PROJECT_DIR}\\SonarLint.Secrets.DotNet.sln
+    PIPELINE_SCRIPTS_DIR: ${PROJECT_DIR}\\pipeline_scripts
+  restore_and_build_script:
+    - source cirrus-env BUILD # sets the $BUILD_NUMBER that is needed for the project version
```
costin-zaharia-sonarsource, line 155: I had to read a bit to see how source works and what's actually doing :). I found this helpful: https://ss64.com/bash/source.html

Do you know where `cirrus-env` is coming from?

sonar-text-dotnet/pipeline_scripts/build_and_restore_solution.ps1
```diff
@@ -0,0 +1,26 @@
+$buildConfiguration="Release"
+
+cd $env:PROJECT_DIR
```
costin-zaharia-sonarsource, line 3: Powershell version would be
```suggestion
Set-Location $env:PROJECT_DIR
```

sonar-text-dotnet/pipeline_scripts/build_and_restore_solution.ps1
```diff
@@ -0,0 +1,26 @@
...
+
+Write-Host Restoring $env:SOLUTION_DIR
+nuget restore -LockedMode -Source "https://pkgs.dev.azure.com/sonarsource/399fb241-ecc7-4802-8697-dcdd01fbb832/_packaging/slvs_input/nuget/v3/index.json"  # nuget version is 6.3.1 https://github.com/SonarSource/re-ci-images/blob/master/ec2-images/build-base-windows-dotnet.pkr.hcl#L39
```
costin-zaharia-sonarsource, line 6: We can extract `$nugetRepository = "https://pkgs.dev.azure.com/sonarsource/399fb241-ecc7-4802-8697-dcdd01fbb832/_packaging/slvs_input/nuget/v3/index.json"` and use the variable instead.

costin-zaharia-sonarsource, line 6: > # nuget version is 6.3.1 https://github.com/SonarSource/re-ci-images/blob/master/ec2-images/build-base-windows-dotnet.pkr.hcl#L39

I think it's useful to link the source but the version can easily get out of sync. We can move the comment above the command and remove the hardcoded version.

sonar-text-dotnet/pipeline_scripts/build_and_restore_solution.ps1
```diff
@@ -0,0 +1,26 @@
...
+
+$signArtifacts=$false
+if(($env:CIRRUS_BRANCH -eq "master" ) -or ($env:CIRRUS_BRANCH.StartsWith("branch-")))
+{
+    $signArtifacts=$true
+    Write-Host "signArtifacts variable was set to {$signArtifacts}".
+}
```
costin-zaharia-sonarsource, line 13: Since this is not yet completed (and not used for real) I would remove the signing-related code from this PR. 

In this way we will be able the full signing logic in a dedicated PR, we will have the full context, and the PR will be easier to understand.

sonar-text-dotnet/pipeline_scripts/run_uts.ps1
```diff
@@ -0,0 +1,4 @@
+$utsDllPath="$env:PROJECT_DIR\\src\\SonarLint.Secrets.DotNet.UnitTests\\bin\\Release\\net48\\SonarLint.Secrets.DotNet.UnitTests.dll"
+
+Write-Host "Running UTs from path: $utsDllPath"
+dotnet test  $utsDllPath  --configuration "Release"  --no-build
```
costin-zaharia-sonarsource, line 4:
```suggestion
dotnet test $utsDllPath --configuration "Release" --no-build
```

Can we maybe run `dotnet test` on the solution instead of dlls? In this way, we will not have to change the script if we decide to change the targeting version or to add more.

sonar-text-dotnet/src/SonarLint.Secrets.DotNet.UnitTests/DotNetOnly/RegexMatcherTest_NoJavaTests.cs
```diff
@@ -35,7 +35,7 @@ public void Find_Simple()
             actual.First().Length.Should().Be(3);
             actual.First().Text.Should().Be("123");
 
-            actual.Last().StartIndex.Should().Be(52);
+            actual.Last().StartIndex.Should().Be(50);
```
costin-zaharia-sonarsource, line 38: That's an interesting change in results. Do you know what change caused the difference?

sonar-text-dotnet/src/SonarLint.Secrets.DotNet.UnitTests/EntropyCheckerTest.cs
```diff
@@ -16,6 +16,8 @@ namespace SonarLint.Secrets.DotNet.UnitTests
     [TestClass]
     public class EntropyCheckerTest
     {
+        private readonly string commonTestFilePath = "..\\sonar-text-plugin\\src\\test\\resources\\EntropyChecker\\";
```
costin-zaharia-sonarsource, line 19: can be a constant.

sonar-text-dotnet/src/SonarLint.Secrets.DotNet.UnitTests/Rules/AlibabaCloudAccessKeySecretsRuleTest.cs
```diff
@@ -68,7 +68,7 @@ public void TestRuleRegexNegative()
         {
             var testSubject = new AlibabaCloudAccessKeySecretsRule();
 
-            var secrets = testSubject.Find(readFileAndNormalize("src/test/files/google-cloud-account-key/GoogleCloudAccountNegative.json", UTF_8));
+            var secrets = testSubject.Find(readFileAndNormalize("..\\sonar-text-plugin\\src\\test\\resources\\checks\\GoogleCloudAccountKeyCheck\\GoogleCloudAccountNegative.json", UTF_8));
```
costin-zaharia-sonarsource, line 71: We can create a constants class for the test projects and add a `Root` constant. Then use the constant in the tests. In this way, next time the root changes, we will only have to change one file.

sonar-text-dotnet/src/SonarLint.Secrets.DotNet/SonarLint.Secrets.DotNet.csproj
```diff
@@ -14,23 +14,22 @@
...
     <!-- The build number and commit id are set by the CI pipeline. -->
-    <BuildNumber>0</BuildNumber>
     <BranchName>not-set</BranchName>
     <CommitId>not-set</CommitId>
-    <AssemblyVersion>$(Version).$(BuildNumber)</AssemblyVersion>
-    <FileVersion>$(Version).$(BuildNumber)</FileVersion>
-    <InformationalVersion>$(Version).$(BuildNumber) Branch: $(BranchName) CommitId: $(CommitId) Configuration: $(Configuration)</InformationalVersion>
+    <AssemblyVersion>$(Version).$(BUILD_NUMBER)</AssemblyVersion>
+    <FileVersion>$(Version).$(BUILD_NUMBER)</FileVersion>
+    <InformationalVersion>$(Version).$(BUILD_NUMBER) Branch: $(CIRRUS_BRANCH) CommitId: $(CIRRUS_CHANGE_IN_REPO) Configuration: Release)</InformationalVersion>
```
costin-zaharia-sonarsource, line 21: There is a variable set `$buildConfiguration` in the PowerShell script but I see that the configuration is hardcoded here.

costin-zaharia-sonarsource (commented):
I've left a couple of comments. Some small improvements and some questions to better understand how things work.

.cirrus.yml
```diff
@@ -132,3 +141,18 @@ promote_task:
     folder: ${CIRRUS_WORKING_DIR}/.m2/repository
   script: cirrus_promote_maven
   cleanup_before_cache_script: cleanup_maven_repository
+
+build_and_release_sonar_secrets_dotnet_task:
+  depends_on:
+    - promote
```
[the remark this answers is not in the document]
mary-georgiou-sonarsource: I've added a card on trello - when we finish for each card that we don't do in this sprint I will open an issue on git. Wdyt?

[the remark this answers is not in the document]
mary-georgiou-sonarsource: Discussed offline: the other tasks fail if this task does not run last.
There's an open card and it's under investigation why this happens.
For this PR it will be left as is.

.cirrus.yml
mary-georgiou-sonarsource: Yes, it's [here](https://github.com/SonarSource/re-ci-images/blob/master/docker/bin/cirrus-env#L28).

.cirrus.yml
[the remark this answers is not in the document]
mary-georgiou-sonarsource: Update: Seems to work. Will update here.

sonar-text-dotnet/src/SonarLint.Secrets.DotNet.UnitTests/DotNetOnly/RegexMatcherTest_NoJavaTests.cs
mary-georgiou-sonarsource: From some tests I did I assume it is the line breaks.
My IDE and the pipeline seem to work with LF (because the tests were failing also in cirrus).
If I remove the line breaks it correctly figures out the starting point of both passwords - when I add the line breaks back and I count them as one character (LF) then the actual start index of the second password is 50. 
I guess with CRLF (2 chars) is +2.

sonar-text-dotnet/src/SonarLint.Secrets.DotNet.UnitTests/Rules/AlibabaCloudAccessKeySecretsRuleTest.cs
mary-georgiou-sonarsource: Maybe it's better to apply this change in another PR after we are done with the pipeline.

WDYT?
In case you agree I will open a git issue.

costin-zaharia-sonarsource: I was thinking to do this as part of PR as we modify all the paths. It looks to me directly related.

.cirrus.yml
costin-zaharia-sonarsource: Cool. Thanks!

sonar-text-dotnet/src/SonarLint.Secrets.DotNet.UnitTests/EntropyCheckerTest.cs
```diff
@@ -16,7 +16,7 @@ namespace SonarLint.Secrets.DotNet.UnitTests
     [TestClass]
     public class EntropyCheckerTest
     {
-        private readonly string commonTestFilePath = "..\\sonar-text-plugin\\src\\test\\resources\\EntropyChecker\\";
+        private const string commonTestFilePath = "..\\sonar-text-plugin\\src\\test\\resources\\EntropyChecker\\";
```
costin-zaharia-sonarsource: I think the constants should start with uppercase.

mary-georgiou-sonarsource: This has been replaced by the universal const in the new class with consts, which has the correct case. :D

sonar-text-dotnet/src/SonarLint.Secrets.DotNet.UnitTests/TestConstants.cs
```diff
@@ -0,0 +1,14 @@
+﻿/*
+ * Copyright (C) 2021-2023 SonarSource SA
+ * All rights reserved
+ * mailto:info AT sonarsource DOT com
+ */
+
+namespace SonarLint.Secrets.DotNet.UnitTests
+{
+    internal static class TestConstants
```
costin-zaharia-sonarsource, line 9: Since the file is located in the Test project I find the prefix redundant. It can be `Constants`.

sonar-text-dotnet/src/SonarLint.Secrets.DotNet.UnitTests/TestConstants.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        public const string RootTestFileDirectory = "..\\sonar-text-plugin\\src\\test\\resources\\";
```
costin-zaharia-sonarsource, line 11: How about `RootPath` instead of `RootTestFileDirectory`.

sonar-text-dotnet/src/SonarLint.Secrets.DotNet.UnitTests/TestConstants.cs
```diff
@@ -0,0 +1,14 @@
...
+
```
costin-zaharia-sonarsource, line 12: pedantic: empty line

.cirrus.yml
```diff
@@ -20,6 +20,15 @@ env:
   # Allows to run builds for the 50 last commits in a branch:
   CIRRUS_CLONE_DEPTH: 1
 
+win_vm_definition: &WINDOWS_VM_DEFINITION
+  ec2_instance:
+    experimental: true # see https://github.com/cirruslabs/cirrus-ci-docs/issues/1051
```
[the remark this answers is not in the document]
costin-zaharia-sonarsource: Did you try to remove it to see if it's still needed?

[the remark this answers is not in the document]
mary-georgiou-sonarsource: 🤦
Trying it now!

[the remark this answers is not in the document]
mary-georgiou-sonarsource: @costin-zaharia-sonarsource Looks like it works without the experimental tag.
I think we can remove it for now (I'll still drop a question to the REs to clarify).

.cirrus.yml
```diff
@@ -132,3 +140,20 @@ promote_task:
     folder: ${CIRRUS_WORKING_DIR}/.m2/repository
   script: cirrus_promote_maven
   cleanup_before_cache_script: cleanup_maven_repository
+
+build_and_release_dotnet_task:
```
csaba-sagi-sonarsource, line 144: Can we rename the task? It seems a bit confusing to me as it does not do the release and if I am not mistaken it will not do it in the future either. As the whole dotnet logic will be part of this task it should somehow reflect that. And at the same time I don't have a really good suggestion how to name it.

mary-georgiou-sonarsource: I can rename it to just `build` but this is a scaffold of the pipeline and it will change again anyways. 
So we might leave it as is for this PR? Unless there's already a plan of how we want to organize the tasks/scripts.

.cirrus.yml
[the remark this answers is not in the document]
mary-georgiou-sonarsource: @costin-zaharia-sonarsource The reply was that the tag is still useful, and we should wait for a feature merge on the cirrus side to remove it.
I asked what is the problem that we might have without it, but until we get an answer, I added it back as it should be needed.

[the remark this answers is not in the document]
mary-georgiou-sonarsource: So this tag is activating experimental feature(s); the one feature that is needed is adding tags on aws; those are required by sonar and RE internal conventions.

costin-zaharia-sonarsource (approved):
LGTM!

.cirrus.yml
csaba-sagi-sonarsource: ok, leave it as is and we will rename when adding the other steps.

PR #89, opened 2023-01-31
Replace sonar-text-dotnet package licence with GNU Lesser General Public License

mary-georgiou-sonarsource:
The project `sonar-secrets-dotnet`  is now opensource and thus the old license restriction is not needed anymore.
When the nuget package is now created it includes the license of the sonar-text root directory (GNU Lesser General Public License).


costin-zaharia-sonarsource (approved):
LGTM!

marco-comi-sonarsource (approved):

PR #92, opened 2023-02-01
Sign the nuget package

csaba-sagi-sonarsource:


csaba-sagi-sonarsource:
It will be implemented in a different PR.

PR #94, opened 2023-02-01
Fail sonar-text-dotnet task if build or restore script fails.

mary-georgiou-sonarsource:
Currently, the build script does not fail if the build fails(under investigation why) and we see it only because the unit test script fails.

Build failing silently before: https://cirrus-ci.com/task/5420589876248576
Build failing loud and clear after this change: https://cirrus-ci.com/task/5962763059593216


sonar-text-dotnet/pipeline_scripts/build_and_restore_solution.ps1
```diff
@@ -17,3 +17,7 @@ dotnet build `
      /p:CommitId=$env:CIRRUS_CHANGE_IN_REPO `
      /p:BranchName=$env:CIRRUS_BRANCH `
      /p:BuildNumber=$env:BUILD_NUMBER
+
+if ($LASTEXITCODE -eq 1) {
```
mary-georgiou-sonarsource, line 21: Any better way to pull the MSBuild exit code?

mary-georgiou-sonarsource: This is how it looks now when it fails:
https://cirrus-ci.com/task/4720181440675840

costin-zaharia-sonarsource (approved):
LGTM!

PR #96, opened 2023-02-02
Sign the assembly and the nuget package

csaba-sagi-sonarsource:
See the logs of a previous run to check that the signing was successful.


sonar-text-dotnet/pipeline_scripts/build_and_pack_solution.ps1
```diff
@@ -0,0 +1,43 @@
+function CheckIfSuccessful($StepName) {
+if(!$?) {
+    Write-Host "sonar-text-dotnet ${StepName} failed."
+    Exit $LASTEXITCODE
+}
+}
+
+Set-Location $env:PROJECT_DIR
+$SIGN_DOTNET_ASSEMBLY = "$CIRRUS_BRANCH" -eq "master" -or "$CIRRUS_BRANCH".startsWith("branch-.*")
```
csaba-sagi-sonarsource, line 9: It is the same logic as in the previous pipeline.

sonar-text-dotnet/pipeline_scripts/build_and_pack_solution.ps1
```diff
@@ -0,0 +1,43 @@
...
+    /p:BuildNumber=$env:BUILD_NUMBER `
+    /p:SignAssembly=$SIGN_DOTNET_ASSEMBLY `
+    /p:AssemblyOriginatorKeyFile=$env:SNK_PATH
+CheckIfSuccessful "build"
+
+Write-Host Packing $env:SOLUTION_DIR
+dotnet pack $env:SOLUTION_DIR -o $env:PROJECT_DIR\artifacts -c $env:BUILD_CONFIGURATION --no-build
+CheckIfSuccessful "packaging"
+
+if($SIGN_DOTNET_ASSEMBLY) {
```
csaba-sagi-sonarsource, line 38: It is the same as in the previous pipeline. And the ITs are run both on master and PR, and they will be green and inconclusive on non-master/feature branch runs.

.cirrus.yml
```diff
@@ -152,13 +152,31 @@ build_and_release_dotnet_task:
...
+    PFX_PATH: ${CIRRUS_WORKING_DIR}\\SonarSource-2021-2023.pfx
+    ENCODED_PFX_CERT: VAULT[development/kv/data/sign/dotnet data.cert_pfx]
+    ENCODED_PFX_CERT_PATH: ${CIRRUS_WORKING_DIR}\\Encoded_SonarSource-2021-2023.pfx
+  sign_cert_pfx_file:
+    path: ${ENCODED_PFX_CERT_PATH}
+    variable_name: ENCODED_PFX_CERT
+  sonarSourceSecret_snk_file:
+    path: ${ENCODED_SONARSOURCE_SNK_PATH} # this is the base64 encoded file
+    variable_name: ENCODED_SONARSOURCE_SNK
+  base64_decode_files_script: |
```
csaba-sagi-sonarsource, line 168: The files in vault are base64 encoded so this step is needed to base64 decode them.

.cirrus.yml
```diff
@@ -152,13 +152,31 @@ build_and_release_dotnet_task:
...
+    base64 --decode ${ENCODED_SONARSOURCE_SNK_PATH} > ${SNK_PATH}
```
csaba-sagi-sonarsource, line 169: Originally I wanted to have only 1 file without encoding in its name and overwriting itself during the decoding, however, I could not do it. I always got an empty file, so I opted for this 2-file solution.

sonar-text-dotnet/pipeline_scripts/build_and_pack_solution.ps1
costin-zaharia-sonarsource: shouldn't be `"$CIRRUS_BRANCH".startsWith("branch-.")` instead? 

I think that the startsWith will read the `*` as it is and not as a pattern.

csaba-sagi-sonarsource: The . is also not needed. Fix is on its way.

costin-zaharia-sonarsource (approved):
LGTM!

PR #98, opened 2023-02-03
Set project version instead of commit id when analysis runs on master.

mary-georgiou-sonarsource:


sonar-text-dotnet/pipeline_scripts/analysis_begin_step.ps1
```diff
@@ -1,12 +1,12 @@
...
 
-if ($env:CIRRUS_PR -eq "") {
+if ($env:CIRRUS_BRANCH -eq "master") {
     Write-Host "Execute analysis begin step for master branch"
 
     SonarScanner.MSBuild.exe begin `
         /k:$env:PROJECT_NAME `
         /n:$env:PROJECT_NAME `
-        /v:$env:CIRRUS_CHANGE_IN_REPO `
+        /v:$env:PROJECT_VERSION `
```
mary-georgiou-sonarsource, line 9: When in master I set the project version instead of the commit as it was before.

sonar-text-dotnet/pipeline_scripts/extract_project_version.ps1
```diff
@@ -0,0 +1,11 @@
+$versionFilePath = "$env:CIRRUS_WORKING_DIR\sonar-text-dotnet\Directory.Build.props"
```
mary-georgiou-sonarsource, line 1: This script is more or less the same with [this one in the azure pipeline](https://github.com/SonarSource/sonar-secrets/blob/master/sonar-secrets-dotnet/azure-pipelines-dotnet.yml#L51).

I could not pull the version from the `csproj` (the `Version` propertyGroup does not exist there and when I pull another property group (for example `AssemblyVersion`) the env variables are not properly transformed to their values).


I pulled the version from the Directory.Build.props which is the file used to set the version in the first place ([see example of version bump in the old repo](https://github.com/SonarSource/sonar-secrets/commit/c417983fb496c1678f60a0d1740bddb71adbf59e)).

sonar-text-dotnet/pipeline_scripts/analysis_begin_step.ps1
```diff
@@ -1,12 +1,12 @@
 Set-Location $env:PROJECT_DIR
 
-if ($env:CIRRUS_PR -eq "") {
+if ($env:CIRRUS_BRANCH -eq "master") {
```
costin-zaharia-sonarsource, line 3: this will again handle only the master. what about the other feature branches?

mary-georgiou-sonarsource: Fixed.

costin-zaharia-sonarsource (approved):
LGTM!

PR #102, opened 2023-02-08
Update the project version before using it

costin-zaharia-sonarsource:
Previously, the variable was set only after the usage, and as a result, the value was not properly sent to SQ.


.cirrus.yml
```diff
@@ -178,7 +178,6 @@ build_and_release_dotnet_task:
     # See https://github.com/SonarSource/re-ci-images/blob/master/docker/bin/cirrus-env#L28
     - source cirrus-env QA
     - PowerShell -NonInteractive -NoProfile -File "${PIPELINE_SCRIPTS_DIR}\\build_and_pack_solution.ps1"
-    - PowerShell -NonInteractive -NoProfile -File "${PIPELINE_SCRIPTS_DIR}\\extract_project_version.ps1"
```
costin-zaharia-sonarsource, line 181: the order was wrong, the variable was set here, after the begin step already run

sonar-text-dotnet/pipeline_scripts/analysis_begin_step.ps1
```diff
@@ -1,3 +1,11 @@
+$versionFilePath = "$env:CIRRUS_WORKING_DIR\sonar-text-dotnet\Directory.Build.props"
```
csaba-sagi-sonarsource, line 1: NITPICK: Why don`t we use project dir and then directly access the build props in it?

csaba-sagi-sonarsource (approved):
LGTM! I left one nitpick comment.

PR #112, opened 2023-05-12
SONARTEXT-10 Migrate rules to educational format

nils-werner-sonarsource:


jonas-wielage-sonarsource (approved):
LGTM! I think this PR would be a good opportunity to update the parent pom version aswell.
Current version is 68.0.0.247

PR #116, opened 2023-05-22
SONARTEXT-20 Implement minimal example using only required fields of the configuration file

jonas-wielage-sonarsource:


sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/configuration/deserialization/SpecificationDeserializationException.java
```diff
@@ -0,0 +1,26 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.secrets.configuration.deserialization;
+
+public class SpecificationDeserializationException extends RuntimeException {
```
nils-werner-sonarsource, line 22: `DeserializationException` would also be okay for me since it could be reused for other deserializer

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/configuration/deserialization/SpecificationDeserializer.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import org.sonar.plugins.secrets.configuration.model.Specification;
+
+public class SpecificationDeserializer {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper(new YAMLFactory());
+
+  private static final Set<String> SPECIFICATIONS = Set.of("minSpec.yaml");
```
nils-werner-sonarsource, line 35: Please provide this within the test method and not in the "production" class.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/configuration/deserialization/SpecificationDeserializer.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private SpecificationDeserializer() {
+  }
+
+  public static List<Specification> deserializeProvidedSpecifications() {
+    ArrayList<Specification> specifications = new ArrayList<>();
+    for (String specification : SPECIFICATIONS) {
+      URL specificationUrl = SpecificationDeserializer.class.getResource(specification);
```
nils-werner-sonarsource, line 43: Please do not relay on the class resources for the deserialization.

nils-werner-sonarsource (approved):
LGTM! Please see my comments about the `minSpec` source. Since it is a test source, we should provide the path within the test class. The deserializer should expect a URL or a Path or even the content of the file.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/configuration/deserialization/SpecificationDeserializer.java
jonas-wielage-sonarsource: @nils-werner-sonarsource As I will be deleting this method, this is no longer relevant.
But for the future, how should I retrieve the resource?

Should I call `Thread.currentThread().getContextClassLoader().getResource(...)` and make the method non-static, or do you have another suggestion?

PR #132, opened 2023-06-02
SONARTEXT-34 Add basic post filter to detection logic

jonas-wielage-sonarsource:


sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/EntropyChecker.java
```diff
@@ -24,14 +24,18 @@
...
-  public static final double ENTROPY_THRESHOLD = 4.2;
+  public static final double DEFAULT_ENTROPY_THRESHOLD = 4.2;
 
   private EntropyChecker() {
     // utility class
   }
 
-  public static boolean hasLowEntropy(String str) {
-    return calculateShannonEntropy(str) < ENTROPY_THRESHOLD;
+  public static boolean hasLowEntropyWithDefaultThreshold(String str) {
```
nils-werner-sonarsource, line 33:
```suggestion
  public static boolean hasLowEntropy(String str) {
```
For me, it would be also clear that a default value is used when not threshold is required.

nils-werner-sonarsource (approved):
LGTM! See nitpick comment.

PR #133, opened 2023-06-02
Update issue management reference in pom

nils-werner-sonarsource:


jonas-wielage-sonarsource (approved):
LGTM!
And TIL that this link will redirect to the current board

PR #135, opened 2023-06-02
SONARTEXT-33 Usage of migrated rules from specification files

jonas-wielage-sonarsource:


sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SecretMatcher.java
nils-werner-sonarsource, line 1: Nice layer!

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SpecificationLoader.java
```diff
@@ -34,18 +35,28 @@
...
-  private final Map<String, Rule> rulesMappedToKey;
+
+  private static final Set<String> DEFAULT_SPECIFICATIONS = Set.of(
+    "alibaba.yaml",
+    "aws.yaml",
+    "azure.yaml",
+    "gcp.yaml",
+    "google-api.yaml",
+    "ibm.yaml",
+    "mws.yaml"
```
nils-werner-sonarsource, line 46: Fine for now. Later we need to collect all files automatically.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SpecificationLoader.java
```diff
@@ -34,18 +35,28 @@
...
   }
 
   public SpecificationLoader(String specificationLocation, Set<String> specifications) {
     rulesMappedToKey = initialize(specificationLocation, specifications);
   }
 
-  private static Map<String, Rule> initialize(String specificationLocation, Set<String> specifications) {
-    Map<String, Rule> keyToRule = new HashMap<>();
+  private static Map<String, List<Rule>> initialize(String specificationLocation, Set<String> specifications) {
+    Map<String, List<Rule>> keyToRule = new HashMap<>();
```
nils-werner-sonarsource, line 59: Why there should be multiple rules to a single key? Same key but multiple rules for the different providers?

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/SecretMatcherAssert.java
```diff
@@ -0,0 +1,64 @@
...
+import org.assertj.core.api.AbstractAssert;
+
+public class SecretMatcherAssert extends AbstractAssert<SecretMatcherAssert, SecretMatcher> {
+
+  static Set<String> strings = Set.of(
+    "candidate secret with low entropy",
+    "candidate secret with low entropy and patternNot:EXAMPLEKEY",
+    "candidate secret with high entropy: lasdij2338f,.q29cm2acasd has patternNot:EXAMPLEKEY",
+    "candidate secret with high entropy: lasdij2338f,.q29cm2acasd"
+  );
```
nils-werner-sonarsource, line 35: I don't get the hardcoded quite specific strings here.

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/SecretMatcherTest.java
```diff
@@ -0,0 +1,74 @@
...
+  @Test
+  void testConstructionOfSimpleDetection() {
+    Rule rule = ReferenceTestModel.constructMinimumSpecification().getProvider().getRules().get(0);
+
+    PatternMatcher patternMatcher = new PatternMatcher("\\b(rule matching pattern)\\b");
+    SecretMatcher expectedMatcher = new SecretMatcher(rule, patternMatcher, s -> true);
+
+    SecretMatcher actualMatcher = SecretMatcher.build(rule);
+
+    SecretMatcherAssert.assertThat(actualMatcher).behavesLike(expectedMatcher);
```
nils-werner-sonarsource, line 40: Please use a static import here.

nils-werner-sonarsource (approved):
LGTM! Please see my comments and answer my questions.

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/SecretMatcherAssert.java
jonas-wielage-sonarsource: @nils-werner-sonarsource As there is no equals for predicates I wanted to test that they behave the same. As I am testing postFilter I put strings in there,which I knew would evaluate to true / false based on the actual filter

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SpecificationLoader.java
jonas-wielage-sonarsource: @nils-werner-sonarsource 
Because of the restructuring of detection, it is now only possible to have a single pattern detection for a rule. So now matchEither for two patterns, only for auxiliaryPatterns.
Because of this and to use different messages for the same ruleKey, it is possible to have two rules in the specification which will raise an issue based on the same key

PR #143, opened 2023-06-27
[APPSEC-788] Add support for databases (Postgres, MySQL, Mongo, ODBC/JDBC)

jamie-anderson-sonarsource:
https://sonarsource.atlassian.net/browse/APPSEC-788

### DIFF
--- a/sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/mongodb.yaml
+++ b/sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/mongodb.yaml
@@ -18,44 +18,44 @@ provider:
           # Look for URIs in the format "protocol://username:password@server:port/..."
           # The protocol is "mongo" or "mongodb" but can also be followed by a subprotocol, e.g. "mongodb+srv".
           # Certain special characters need to be percent-encoded so we can break matching when we find them.
-          pattern: "\\bmongo(?:db)?(?:\\+\\w+)?://[^:@/ ]:([^@/ ]+)@"
+          pattern: "\\bmongo(?:db)?(?:\\+\\w+)?://[^:@/ ]+:([^@/ ]+)@"
         post:
           # Avoid matching values found on SourceGraph that look like dummy passwords or insertions.
-          patternNot: "(?i)^(\\$[a-z_]*)?(db|my)?_?pass(word|wd)?|^\\[[\\w\\t \\-]+\\]$|^<[\\w\\t \\-]+>$|^%s$|^\\$\\w+$|^\\{+[^}]*\\}+$|^\\$\\(\\w+\\)$"
+          patternNot: "(?i)^(\\$[a-z_]*)?(db|my)?_?pass(word|wd)?|^\\[[\\w\\t \\-]+\\]$|^<[\\w\\t \\-]+>$|^%s$|^\\$\\w+$|^\\$?\\{+[^}]*\\}+$|^\\$\\(\\w+\\)$"
       examples:
-        - text: >
+        - text: |
             backend "pg" {
               conn_str = "mongo://user:pass@database.example/terraform_backend"
             }
           containsSecret: false
-        - text: >
+        - text: |
             connStr = fmt.Sprintf("mongodb://%s:%s@:%s/%s%ssslmode=%s&host=%s",
               url.PathEscape(dbUser), url.PathEscape(dbPasswd), port, dbName, dbParam, dbsslMode, host)
           containsSecret: false
-        - text: >
+        - text: |
             if (!newConfig.DATABASE_URL) {
               const encodedUser = encodeURIComponent(newConfig.POSTHOG_DB_USER)
               const encodedPassword = encodeURIComponent(newConfig.POSTHOG_DB_PASSWORD)
               newConfig.DATABASE_URL = `mongodb://${encodedUser}:${encodedPassword}@${newConfig.POSTHOG_MONGO_HOST}:${newConfig.POSTHOG_MONGO_PORT}/${newConfig.POSTHOG_DB_NAME}`
             }
           containsSecret: false
-        - text: >
+        - text: |
             SqlSettingsDefaultDataSource = "mongodb+srv://mmuser:mostest@localhost/mattermost_test?sslmode=disable&connect_timeout=10&binary_parameters=yes"
           containsSecret: true
           match: mostest
-        - text: >
+        - text: |
             passwordFile := fs.String("password", "../../../tools/secrets/password.txt", "password file")
             databasePrefix := fs.String("database-prefix", "mongodb://mongo:mongo_password_padded_for_security@localhost:27017/ocr2vrf-test", "database prefix")
             databaseSuffixes := fs.String("database-suffixes", "sslmode=disable", "database parameters to be added")
           containsSecret: true
           match: mongo_password_padded_for_security
-        - text: >
+        - text: |
             driver: mongodb
             dsn: mongodb://foouser:foopass@localhost:27017/testdb?sslmode=disable
             table: footable
           containsSecret: true
           match: foopass
-        - text: >
+        - text: |
             AIRFLOW_CONN_METADATA_DB=mongodb+srv://airflow:airflow@mongo:27017/airflow
             AIRFLOW_VAR__METADATA_DB_SCHEMA=airflow
           containsSecret: true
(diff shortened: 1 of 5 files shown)


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/odbc.yaml
```diff
@@ -0,0 +1,204 @@
...
+            content:
+              - datasource
+        matching:
+          # Same as the file java-datasource-unquoted, except that the value is surrounded by double quotes.
+          # This starts with a negative lookbehind to see if there's a double quote shortly before the string. This is
+          # to prevent false negative matches with code like:
+          #   param += " --spring.datasource.url=" + url
+          #         + " --spring.datasource.username=" + username
+          #         + " --spring.datasource.password=" + password;
+          pattern: "(?<!\"[^\"\\r\\n]{0,6})\\b[0-9a-z_\\-.]+\\.datasource(?:\\.[0-9a-z_\\-.]+)*\\.password[ \\t]*=[ \\t]*\"((?:\\\\\"|[^\"])+?)\""
```
gaetan-ferry-sonarsource, line 72: I guess the same will apply.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/odbc.yaml
```diff
@@ -0,0 +1,204 @@
...
+          context:
+            - patternAround: "\\b[0-9a-z_\\-.]+\\.datasource(?:\\.[0-9a-z_\\-.]+)*\\.username[ \\t]*="
```
gaetan-ferry-sonarsource, line 74: same propery style problem.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/odbc.yaml
```diff
@@ -0,0 +1,204 @@
...
+    - id: jdbc-connection-string
+      metadata:
+        name: Database passwords should not be disclosed
+      detection:
+        pre:
+          include:
+            content:
+              - jdbc
+          reject:
+            ext:
```
gaetan-ferry-sonarsource, line 108: This does not pass the schema validation.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/odbc.yaml
```diff
@@ -0,0 +1,204 @@
...
+              - .adoc
+              - .example
+              - .html
+              - .md
+              - .mdx
+              - .template
+        matching:
+          # JDBC URIs are in the form "jdbc:protocol://server:port/database?user=...&password=...&paramN=..."
+          pattern: "\\bjdbc:[^\"\\s?]+\\?(?:[^&\"\\s]+&)*password=([^&\"\\s]+)(?:[&\"\\s])"
```
gaetan-ferry-sonarsource, line 117: I feel like this pattern will overlap the mysql one in a lot of cases. Other than that LGTM

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/odbc.yaml
```diff
@@ -0,0 +1,204 @@
...
+              "dbtable" -> "departments")).collect().foreach(println)
+          containsSecret: true
+          match: P@ssw0rd
+
+    - id: odbc-connection-string
+      metadata:
+        name: Database passwords should not be disclosed
+      detection:
+        pre:
+          reject:
```
gaetan-ferry-sonarsource, line 154: Does not pass the validation.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/postgresql.yaml
```diff
@@ -0,0 +1,294 @@
...
+    - id: postgres-env-pg-password-quoted
+      metadata:
+        name: PostgreSQL database passwords should not be disclosed
+      detection:
+        pre:
+          include:
+            content:
+              - PG_PASSWORD
+          reject:
+            ext:
```
gaetan-ferry-sonarsource, line 180: blah

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/postgresql.yaml
```diff
@@ -0,0 +1,294 @@
...
+    - id: psql-cli-quoted
+      metadata:
+        name: PostgreSQL database passwords should not be disclosed
+      detection:
+        pre:
+          include:
+            content:
+              - psql
+          reject:
+            ext:
```
gaetan-ferry-sonarsource, line 271: :'|

gaetan-ferry-sonarsource (changes requested):
A "few" things to look into. See comments for details.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/mongodb.yaml
```diff
@@ -0,0 +1,118 @@
...
+    - id: mongodb-cli-unquoted
+      metadata:
+        name: MongoDB database passwords should not be disclosed
+      detection:
+        pre:
+          include:
+            content:
+              - mongodb
+          reject:
+            ext:
```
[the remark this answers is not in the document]
jamie-anderson-sonarsource: It appears that the implementation here differs from what's in the specification. Only the `paths` module is currently allowed under the `reject` heading.

I've pointed this out [in #mmf-3132 in Slack](https://sonarsource.slack.com/archives/C0585H9EPCZ/p1688030819426969), so I'm going to leave the implementation in place for now. If it causes problems, we can convert the extensions to patterns for the `paths` module.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/mongodb.yaml
```diff
@@ -0,0 +1,118 @@
...
+              - .md
+              - .html
+              - .example
+              - .template
+        matching:
+          pattern: "\\bmongodb\\b.{1,50}--password\\s+(?![\"'])([^\\s]+)"
```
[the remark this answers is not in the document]
jamie-anderson-sonarsource: You're right about `mongodb`. I found a web page with bad info and got lots of matches because of `mongodb://` URLs.

It turns out that there are several tools that start with `mongo` that use the same syntax. The updated regex will support all of them.

I've also discovered that `-p` is an alias for `--password` and will need a small negative lookbehind to make that work correctly. The negative lookahead near the capture group will be removed.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/mysql.yaml
```diff
@@ -0,0 +1,281 @@
...
+            content:
+              - MYSQL_PASSWORD
+          reject:
+            ext:
+              - .md
+              - .html
+              - .example
+              - .template
+        matching:
+          pattern: "\\bMYSQL_PASSWORD=(?!\\\\?[\"'])([^\\s]+)(?:$|\\s)"
```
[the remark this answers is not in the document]
jamie-anderson-sonarsource: I spotted this use case on SourceGraph:
```shell
remote_cmd $h "sed -i \"s/^MYSQL_PASSWORD=.*/MYSQL_PASSWORD=\'$new_password\'/\" $MYSQLCHK "
```

It's the only example of `\'` I could find, but it's easy to support and should raise very few false negatives.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/mysql.yaml
```diff
@@ -0,0 +1,281 @@
...
+              - mysql
+              - mysqladmin
+          reject:
+            ext:
+              - .md
+              - .html
+              - .example
+              - .template
+        matching:
+          pattern: "\\bmysql(?:admin)?\\b.{1,50}(?<!\")(?:-p|--password=)(?!\")([^\\s\"]+)"
```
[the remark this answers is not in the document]
jamie-anderson-sonarsource: Many of the suggestions here will be implemented.

I've put the 3 character minimum on all command lines that take a password parameter. It seems highly unlikely that any "real" password will only use 2 characters. This also has the benefit of not matching `%s` as a password.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/odbc.yaml
```diff
@@ -0,0 +1,204 @@
...
+        name: Database passwords should not be disclosed
+      detection:
+        pre:
+          include:
+            content:
+              - datasource
+          # Match `*.datasource.*.password=some-password` (where the second `*.` is optional).
+          # There also needs to be a `*.datasource.*.username=` nearby.
+        matching:
+          pattern: "\\b[0-9a-z_\\-.]+\\.datasource(?:\\.[0-9a-z_\\-.]+)*\\.password[ \\t]*=[ \\t]*(?!\")([^ \\t]+)[ \\t]*(?:(?<\\s)#.+)?$"
```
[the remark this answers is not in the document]
jamie-anderson-sonarsource: This regex has had a lot of rework and should behave a lot better now.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/odbc.yaml
jamie-anderson-sonarsource: There shouldn't be overlap because the other URI-like patterns expect `protocol://username:password@host:port` whereas this requires the username and password to be passed as query parameters.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/mysql.yaml
```diff
@@ -0,0 +1,287 @@
...
+            ext:
+              - .md
+              - .html
+              - .example
+              - .template
+        matching:
+          pattern: "\\bMYSQL_PASSWORD=(?!\\\\?[\"'])([^\\s]+)(?:$|\\s)"
+        post:
+          # Avoid matching values found on SourceGraph that look like dummy passwords or insertions.
+          patternNot: "^(db|my)?_?pass(word|wd)?$|^None$|^\\*{3,}$|^\\$[A-Z_]+PASS(WORD)?$|^\\$[{(]|^`[a-z]+|^process\\.env\\.|^<[a-z ]{1,10}>?$"
```
gaetan-ferry-sonarsource, line 86: I have the feeling that this would make sense in most of the rules we write. It is used pretty much everywhere in your PR so:

- You could consider moving that at the provider level.
- We could move this to a common configuration. I don't think this is implemented for now tho.
- We could split this big pattern into multi sub things and specific heuristic post filtering module checks.

To clarify, this is just food for thought, I don't expect a change here.

PR #145, opened 2023-06-30
[Secrets] Implement top 50 secrets patterns

gaetan-ferry-sonarsource:
Related RSPEC pull request: https://github.com/SonarSource/rspec/pull/2276


jamie-anderson-sonarsource:
There ended up being 3 separate copies of some bad test data for Google Cloud. This data contains a supposed private key, which is clearly encoded using base64. However, it is littered with `=` characters and they're only used as padding at the very end of the data. These have all been replaced with `0`, and the final character has been replaced with `=`.

jamie-anderson-sonarsource (approved):
I'm happy with the changes.

NOTE: [SONARTEXT-47](https://sonarsource.atlassian.net/browse/SONARTEXT-47) will need to be implemented before these configs work correctly, due to some of them having `reject` entries with file extensions.

PR #149, opened 2023-07-18
SONARTEXT-50 Load and validate schema for all new specified secrets

jonas-wielage-sonarsource:


sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SpecificationLoader.java
```diff
@@ -63,17 +56,32 @@ private static Map<String, List<Rule>> initialize(String specificationLocation,
...
+    // TODO: SONARTEXT-31: Add Context to detection logic
+    Matching matching = rule.getDetection().getMatching();
+    boolean contextIsSet = matching != null && matching.getContext() != null;
+
+    // TODO: SONARTEXT-48: Extend Post-Filter to include heuristic Filter
+    PostModule post = rule.getDetection().getPost();
+    boolean heuristicFilterIsSet = post != null && post.getHeuristicFilter() != null;
+
+    return contextIsSet || heuristicFilterIsSet;
+  }
```
jonas-wielage-sonarsource, line 83: I changed this behavior to only disable tests based on those two missing features. 
Pre-Filter and StatisticalFilter.inputString only server to disable FP's. IMO it doesn't make sense to remove functionality in between releases.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/SecretsRulesDefinition.java
```diff
@@ -49,15 +48,14 @@ public DefaultQualityProfile() {
   }
 
   public static List<Class<?>> checks() {
-    List<Class<?>> checkList = new ArrayList<>(GeneratedCheckList.generatedChecks());
```
petertrr, line 52: Do we still plan to use `GeneratedCheckList`?

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SpecificationLoader.java
```diff
@@ -63,17 +56,32 @@ private static Map<String, List<Rule>> initialize(String specificationLocation,
       try {
         specification = loadSpecification(specificationLocation, specificationFileName);
       } catch (DeserializationException | SchemaValidationException e) {
-        LOG.error(String.format("Could not load specification from file: %s", specificationFileName), e);
+        LOG.error(String.format("%s: Could not load specification from file: %s", e.getClass().getSimpleName(), specificationFileName));
```
petertrr, line 59: Will it work like this?
```suggestion
        LOG.error("{}: Could not load specification from file: {}", e.getClass().getSimpleName(), specificationFileName);
```

sonar-text-plugin/src/test/resources/secretsConfiguration/postFilter/postFilterSpec.yaml
```diff
@@ -0,0 +1,50 @@
...
+          - description: rule reference
+            link: https://docs.aws.amazon.com/IAM/...
+            type: Standards
+      detection:
+        matching:
+          pattern: "\\b(rule matching pattern)\\b"
+        post:
+          statisticalFilter:
+            threshold: 4.2
+          patternNot: EXAMPLEKEY
```
petertrr, line 44: Maybe add other fields here as well (`input_string`, `heuristics`) as they are already present in the model?

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/checks/AwsCheckTest.java
```diff
@@ -31,6 +31,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.sonar.plugins.common.TestUtils.analyze;
 
+
```
petertrr, line 34: Do we need a ticket to add code style checking or spotless plugin?

petertrr (approved):
LGTM! See some of my questions below

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/checks/AwsCheckTest.java
jonas-wielage-sonarsource: Spotless would be good. Let's create a ticket for it.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/SecretsRulesDefinition.java
jonas-wielage-sonarsource: I don't believe so. Should be ideated after the sprint.

PR #150, opened 2023-07-18
SONARTEXT-48 Extend Post-Filter to include heuristic Filter

petertrr:


sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/HeuristicsTest.java
```diff
@@ -0,0 +1,63 @@
...
+class HeuristicsTest {
+    @ParameterizedTest
+    @CsvSource({
+            "/home/user, true",
+            "/home/username, false",
+            "C:\\Users\\User, true",
+            "asecretstr/ng, false"
+    })
+    void shouldDeterminePath(String input, boolean isPath) {
+        Assertions.assertThat(Heuristics.isPath(input)).isEqualTo(isPath);
```
jonas-wielage-sonarsource, line 36: Should be a static import

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/Heuristics.java
jonas-wielage-sonarsource, line 1: Please format this class, as there are 8 code smells related to it.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/Heuristics.java
```diff
@@ -0,0 +1,54 @@
...
+                default:
+                    LOG.warn("Heuristic with the name `{}` is not supported", h);
+                    return false;
+            }
+        });
+    }
+
+    public static boolean isPath(String input) {
+        long fileSeparatorCount = input.chars().filter(c -> c == '/' || c == '\\').count();
+        return fileSeparatorCount >= 2 && fileSeparatorCount * 1. / input.length() > 0.15;
```
jonas-wielage-sonarsource, line 48: `0.15` and `2` should be variables at top level. Also please use parenthesis to make calculation more readable.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/Heuristics.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final Logger LOG = Loggers.get(Heuristics.class);
+
+    public static boolean matchesHeuristics(String candidateSecret, List<String> heuristics) {
+        return heuristics.stream().anyMatch(h -> {
+            switch (h) {
+                case "path": return isPath(candidateSecret);
+                case "uri": return isUri(candidateSecret);
+                default:
+                    LOG.warn("Heuristic with the name `{}` is not supported", h);
+                    return false;
```
jonas-wielage-sonarsource, line 41: Good default

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/Heuristics.java
```diff
@@ -0,0 +1,54 @@
...
+public class Heuristics {
+    private Heuristics() {}
+
+    private static final Pattern uriPattern = Pattern.compile("^(https?|ftps?|file|smtp|imap)://.*$");
+
+    private static final Logger LOG = Loggers.get(Heuristics.class);
+
+    public static boolean matchesHeuristics(String candidateSecret, List<String> heuristics) {
+        return heuristics.stream().anyMatch(h -> {
+            switch (h) {
```
jonas-wielage-sonarsource, line 36: We don't use single character variable names ;)
Also I like if else more, but should be fine.

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/HeuristicsTest.java
```diff
@@ -0,0 +1,63 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.secrets.api;
+
+import java.util.Arrays;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+class HeuristicsTest {
```
jonas-wielage-sonarsource, line 27: There should be a test when the `default` path is taken.

jonas-wielage-sonarsource (approved):
LGTM! Please see my suggestions. Also I would like a test in postFilterSpec.yaml to show that it really works based on the heuristics.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/Heuristics.java
petertrr: Just thought that there might be more heuristics in the future, and `switch` would make more sense in that case

jonas-wielage-sonarsource: Sounds good!

PR #151, opened 2023-07-18
SONARTEXT-32 Add Pre-Filter to detection logic

petertrr:


jonas-wielage-sonarsource (commented):
Here are a few suggestions, I will handle the rest of the review tomorrow.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/PreFilterFactory.java
```diff
@@ -0,0 +1,97 @@
...
+
+  public static Predicate<InputFileContext> createPredicate(@Nullable PreModule pre) {
+    if (pre == null) {
+      return INCLUDE_ALL_FILES;
+    }
+
+    Predicate<InputFileContext> predicate = INCLUDE_ALL_FILES;
+    FileFilter include = pre.getInclude();
+    FileFilter reject = pre.getReject();
+    if (reject != null && include == null) {
```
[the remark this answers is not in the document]
jonas-wielage-sonarsource: And should represent the logic that we aim for.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/PreFilterFactory.java
```diff
@@ -0,0 +1,97 @@
...
+    return anyMatch(filter.getPaths(), PreFilterFactory::matchesPath, ctx) ||
+      anyMatch(filter.getExt(), PreFilterFactory::matchesExt, ctx) ||
+      anyMatch(filter.getContent(), PreFilterFactory::matchesContent, ctx);
+  }
+
+  private static boolean anyMatch(@Nullable List<String> input, BiPredicate<String, InputFileContext> predicate, InputFileContext ctx) {
+    if (input == null) {
+      return false;
+    }
+    return input.stream().anyMatch(s -> predicate.test(s, ctx));
```
jonas-wielage-sonarsource, line 69: I like this approach!
Nonetheless I would change the naming.
`input` -> `filterKeyword` or `filterElements` or something like that.
`predicate` -> `filterFunction`

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SecretMatcher.java
```diff
@@ -22,24 +22,29 @@
...
 
   private final Rule rule;
   private final PatternMatcher patternMatcher;
-  private final Predicate<String> postFilter;
 
-  SecretMatcher(Rule rule, PatternMatcher patternMatcher, Predicate<String> postFilter) {
+  private final Predicate<InputFileContext> preFilter;
+
+  private final Predicate<String> postFilter;
+  SecretMatcher(Rule rule, PatternMatcher patternMatcher, Predicate<InputFileContext> preFilter, Predicate<String> postFilter) {
```
jonas-wielage-sonarsource, line 36: Format is not correct.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SpecificationBasedCheck.java
```diff
@@ -59,6 +59,10 @@ public void analyze(InputFileContext ctx) {
     List<TextRange> foundSecrets = new ArrayList<>();
 
     for (SecretMatcher secretMatcher : matcher) {
+      if (!secretMatcher.getPreFilter().test(ctx)) {
+        continue;
+      }
```
jonas-wielage-sonarsource, line 64: IMO this logic should reside in the secretMatcher. Maybe we should change `secretMatcher.findIn(...)` to accept InputFileContext instead of the content

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/PreFilterFactoryTest.java
```diff
@@ -0,0 +1,156 @@
...
+  @CsvSource({
+          ".doc, file.doc, true",
+          ".doc, file.cpp, false",
+          "pp, file.cpp, true",
+          "'', file.cpp, false",
+  })
+  void testMatchesExt(String ext, String filename, boolean shouldMatch) {
+    InputFileContext ctx = Mockito.mock(InputFileContext.class);
+    Mockito.when(ctx.getInputFile()).thenReturn(Mockito.mock(InputFile.class));
+    Mockito.when(ctx.getInputFile().filename()).thenReturn(filename);
```
jonas-wielage-sonarsource, line 54: Please use static imports.

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/PreFilterFactoryTest.java
```diff
@@ -0,0 +1,156 @@
...
+    assertThat(PreFilterFactory.matchesExt(ext, ctx)).isEqualTo(shouldMatch);
+  }
+
+  @ParameterizedTest
+  @CsvSource({
+          "src/*.cpp, /home/user/project/src/file.cpp, true",
+          "file.cpp, /home/user/project/src/file.cpp, true",
+          "*/file.cpp, /home/user/project/src/file.cpp, true",
+          "'', /home/user/project/src/file.cpp, false",
```
jonas-wielage-sonarsource, line 63: I would like a testcase with a `*` at the end, and failing assertions with a `*`

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/PreFilterFactoryTest.java
```diff
@@ -0,0 +1,156 @@
...
+  })
+  void testMatchesPath(String pathPattern, String filePath, boolean shouldMatch) throws URISyntaxException {
+    InputFileContext ctx = Mockito.mock(InputFileContext.class);
+    Mockito.when(ctx.getInputFile()).thenReturn(Mockito.mock(InputFile.class));
+    Mockito.when(ctx.getInputFile().uri()).thenReturn(new URI("file://" + filePath));
```
jonas-wielage-sonarsource, line 68: We need actual tests for this, as I'm not sure how we will receive the URI.

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/PreFilterFactoryTest.java
```diff
@@ -0,0 +1,156 @@
...
+    Mockito.when(ctx.getInputFile()).thenReturn(Mockito.mock(InputFile.class));
+    Mockito.when(ctx.getInputFile().filename()).thenReturn(filename);
+    Mockito.when(ctx.getInputFile().uri()).thenReturn(new URI("file://" + filename));
+    Mockito.when(ctx.lines()).thenReturn(List.of());
+
+    assertThat(predicate.test(ctx)).isEqualTo(shouldMatch);
+  }
+
+  static Stream<Arguments> inputs() {
+    return Stream.of(
```
jonas-wielage-sonarsource, line 101: There is some inconsistent formatting.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/PreFilterFactory.java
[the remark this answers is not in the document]
petertrr: I believe you're right, thanks for the suggestion

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/SecretMatcherAssert.java
```diff
@@ -47,6 +47,7 @@ public static SecretMatcherAssert assertThat(SecretMatcher actual) {
   public SecretMatcherAssert behavesLike(SecretMatcher expectedMatcher) {
     isNotNull();
     usingRecursiveComparison()
+      .ignoringFields("preFilter")
```
jonas-wielage-sonarsource, line 50: We shouldn't simply ignore it, but write test cases for it aswell.

jonas-wielage-sonarsource (changes requested):
Please see my suggestions. Also I would like an additional integration test. With this we could have a test case with file endings.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/PreFilterFactory.java
```diff
@@ -0,0 +1,97 @@
...
+      return ctx.getInputFile().uri().toString().contains(path);
+    }
+  }
+
+  static boolean matchesContent(String content, InputFileContext ctx) {
+    if (content.isBlank()) {
+      return false;
+    }
+    String contentLowerCase = content.toLowerCase(Locale.getDefault());
+    return ctx.lines().stream().anyMatch(line -> line.toLowerCase(Locale.getDefault()).contains(contentLowerCase));
```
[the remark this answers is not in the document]
petertrr: Yes, this was confirmed with AppSec

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/PreFilterFactory.java
```diff
@@ -0,0 +1,97 @@
...
+  private static boolean anyMatch(@Nullable List<String> input, BiPredicate<String, InputFileContext> predicate, InputFileContext ctx) {
+    if (input == null) {
+      return false;
+    }
+    return input.stream().anyMatch(s -> predicate.test(s, ctx));
+  }
+
+  static boolean matchesPath(String path, InputFileContext ctx) {
+    if (path.isBlank()) {
+      return false;
```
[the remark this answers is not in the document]
petertrr: I believe so. Should we log a warning in this case?

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/PreFilterFactoryTest.java
petertrr: The underlying implementation delegates to `Path#toUri`, so I believe it would be something like this

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/SecretMatcherAssert.java
petertrr: I'm not sure if it makes a lot of sense to also test predicates' behavior here since we already have tests for the predicate building.

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/PreFilterFactory.java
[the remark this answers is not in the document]
jonas-wielage-sonarsource: Sounds good!

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/api/PreFilterFactoryTest.java
petertrr: Apprantly, docs they the following about URI:
``
The only guarantee is that it is unique in the project. You should not assume it is a file:// URI.
```
And all accessors to the path like `absolutePath()` are deprecated. Not sure, how it should be handled. I can use `absolutePath()` for now.

petertrr: After discussion, moved to patching based on globs and `FilePredicates` provided by `sonar-plugin-api`

jonas-wielage-sonarsource (approved):
LGTM, only a small nitpick :)

PR #157, opened 2023-07-20
SONARTEXT-55 Prevent raising multiple issues on the same secret

jonas-wielage-sonarsource:


sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SpecificationBasedCheck.java
```diff
@@ -56,18 +62,20 @@ public void initialize(SpecificationLoader loader) {
...
-          }
-        });
+        .forEach(textRange -> reportIfNoOverlappingSecretAlreadyFound(ctx, textRange, secretMatcher));
+    }
+  }
+
+  public void reportIfNoOverlappingSecretAlreadyFound(InputFileContext ctx, TextRange foundSecret, SecretMatcher secretMatcher) {
+    List<TextRange> reportedSecrets = reportedIssuesForCtx.compute(ctx, (k, v) -> (v == null) ? new ArrayList<>() : v);
+
+    boolean overlappingSecrets = reportedSecrets.stream().noneMatch(reportedSecret -> reportedSecret.overlap(foundSecret));
```
petertrr, line 75: `noOverlappingSecrets` or something similar?

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextAndSecretsSensor.java
```diff
@@ -175,9 +179,10 @@ protected List<Check> getActiveChecks() {
 
   protected void initializeSpecificationBasedChecks(List<Check> checks) {
     SpecificationLoader specificationLoader = new SpecificationLoader();
+    Map<InputFileContext, List<TextRange>> reportedIssuesForCtx = new HashMap<>();
```
petertrr, line 182: I can see that we perform active checks sequentially, but can it be possible that we change it to parallel and would need a `ConcurrentHashMap` here?

petertrr (approved):
LGTM! See my notes below

PR #158, opened 2023-07-20
SONARTEXT-8 [S6290] FP on example keys

petertrr:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/aws.yaml
```diff
@@ -55,6 +55,9 @@ provider:
             AWS_FECRET_KEY=kHeUAwnSUizTWpSbyGAz4f+As5LshPIjvtpswqGb'
             AWS_SECRET_KEY=EXAMPLEKEYCXCgDCUbJq1h7CKwNqnpA1il4MXL+y
           containsSecret: false
+        - text: |
+            c.S3SecretAccessKey = c.Get("PRECISE_CODE_INTEL_UPLOAD_AWS_SECRET_ACCESS_KEY", "wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY", "An AWS secret key associated with a user with access to S3.")
```
jonas-wielage-sonarsource, line 59: I only see the `secret_access_key` added as an example.
The ticket also mentions the `aws_access_key_id="AKIAIOSFODNN7EXAMPLE"`, so please add it as an additional example.
Should belong to the rule with the id `id: aws-access-key-ids`

jonas-wielage-sonarsource (approved):
LGTM, please see my comment before merging

PR #178, opened 2023-08-10
SONARTEXT-57 Update Sonar-Plugin-API to Version 10.0.0.695

jonas-wielage-sonarsource:


pom.xml
```diff
@@ -184,6 +191,12 @@
         <version>${mockito.version}</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-api</artifactId>
+        <version>${version.slf4j}</version>
+        <scope>test</scope>
```
petertrr, line 198: I think this dependency shouldn't be set to `test` in `dependencyManagement` because in child modules we are actually using it in the main source set. Scope is overridden and everything works, but I think without scope here it would be more clear.

petertrr (approved):
LGTM!

PR #185, opened 2023-08-10
SONARTEXT-68 Use Java 17 to build project

petertrr:
* Update properties in pom.xml
* Update requested images on Cirrus
* Use newer scanner in tests


jonas-wielage-sonarsource:
Will be continued in SonarSource/sonar-text#17

PR #196, opened 2023-08-16
SONARTEXT-66 Update generation process for new rules to remove checks with keys that aren't in use anymore

jonas-wielage-sonarsource:


secretSpecificationInclusionGenerator.sh
```diff
@@ -0,0 +1,44 @@
...
+	echo "--------- Generating rspec files for: $line ---------"
+	java -jar ../$ruleapifilename generate -rule $line
+	echo ""
+done < "$input"
+
+rm ../sonar-text-plugin/src/test/resources/templates/rspecKeysToUpdate.txt
+
+echo ""
+echo "---------Finished Generation process---------"
+echo "If the build process fails, this could be due to the enfore-plugin."
```
petertrr, line 43:
```suggestion
echo "If the build process fails, this could be due to the enforcer-plugin."
```

secretSpecificationInclusionGenerator.sh
```diff
@@ -0,0 +1,44 @@
...
+else
+    ruleapifilename="rule-api-snap.jar"
+fi
+
+echo ""
+echo "--------- Generation of Java classes ---------"
+echo ""
+mvn test -Dtest=UpdatingSpecificationFilesGenerator#firstStep test
+
+mvn test -Dtest=UpdatingSpecificationFilesGenerator#secondStep test
```
petertrr, line 14: Is `test` duplicated here?

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/utils/UpdatingSpecificationFilesGenerator.java
```diff
@@ -174,9 +179,38 @@ private void writeSpecificationFileDefinition() {
...
+      String filename = rspecKeysMappedToCheckNames.get(key);
+
+      if (filename != null) {
+        removeUnusedCheck(filename, key);
+      }
+    }
+  }
+
+  private void removeUnusedCheck(String checkName, String rspecKey) {
+    Path checkPath = Path.of("src", "main", "java", "org", "sonar", "plugins", "secrets", "checks", checkName + ".java");
```
petertrr, line 193: Should we make these paths class-level constants?

secretSpecificationInclusionGenerator.sh
```diff
@@ -0,0 +1,44 @@
...
+mvn license:format
+
+echo ""
+echo "---------Formatting code---------"
+echo ""
+mvn spotless:apply
+
+cd sonarpedia-secrets
+
+# generated file where rspec keys to update are stored
```
petertrr, line 28: I suggest to specify also, at which stage it is generated.

petertrr (approved):
LGTM, see my suggestions

PR #201, opened 2023-08-22
APPSEC-1004: Add Firebase legacy FCM keys

loris-s-sonarsource:


daniel-teuchert-sonarsource (approved):
LGTM!
Still needs the html and json files generated via rspec though.

PR #203, opened 2023-08-24
APPSEC-1003: S6718 Add stripe webhook url

daniel-teuchert-sonarsource:


loris-s-sonarsource (approved):
LGTM ! I would suggest adding one test related to only one of the post module strings.

IMPORTANT: Do not merge now ! Please generate the HTML files and re run tests before ! Plus make sure tests are green

PR #206, opened 2023-08-24
S6687 Improved django secret detection

daniel-teuchert-sonarsource:


loris-s-sonarsource (commented):
- regex looks good
-  the patternNot regexes are not that complex so multilining them might be overkill but it's a plus so let's keep it
- In the last petternNot there's "\\{\\{.*?}}|\" and I am not sure why it's useful. Maybe for templating. However it does not take actual bash var envs, do you think that would make sense ? I really don't know
- last thing: Did you test this ?

loris-s-sonarsource:
Btw, a user also have this kind of situation:

```python
??? = os.getenv(
  "XXXX", "JNnGGT-cJAZWfFWLIveoTAxMj6RuAe1iWxAAF4IA"
)
```

I think we can raise something here too, on the default value

daniel-teuchert-sonarsource:
> * regex looks good
> * the patternNot regexes are not that complex so multilining them might be overkill but it's a plus so let's keep it
> * In the last petternNot there's "\{\{.*?}}|" and I am not sure why it's useful. Maybe for templating. However it does not take actual bash var envs, do you think that would make sense ? I really don't know
> * last thing: Did you test this ?

There is a FP example like this:
`SECRET_KEY={{ random_ascii_string(60) }} `

What kind of tests are you expecting me to run?
I tested with the test cases and did a quick sanity check via sourcegraph.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/django.yaml
```diff
@@ -66,7 +96,13 @@ provider:
           heuristicFilter:
             heuristics:
               - "path"
-          patternNot: "(?i)(\\{\\{.*?}}|sample|example|x{4,}|0{4,})"
+          patternNot:
```
loris-s-sonarsource, line 99:
```suggestion
          # Some FPs include templating, such as SECRET_KEY={{ random_ascii_string(60) }}
          patternNot:
```

loris-s-sonarsource (approved):
LGTM

PR #207, opened 2023-08-25
S6698: Improved postgresql secret detection precision

daniel-teuchert-sonarsource:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/postgresql.yaml
```diff
@@ -30,7 +30,18 @@ provider:
           # Avoid matching values found on SourceGraph that look like dummy passwords or insertions.
           # Note: Anything matching values like "${DB_PASSWORD:-postgres}" will be ignored and processed by another
           # rule.
-          patternNot: "(?i)^(\\$[a-z_]*)?(db|my)?_?pass(word|wd)?|^\\[[\\w\\t \\-]+\\]$|^<[\\w\\t \\-]+>$|^%s$|^\\{+[^}]*\\}+$|^\\$\\(\\w+\\)$|\\$\\{\\w+(:-\\w+)?\\}"
+          # The last line was added to this pattern to sort out any code that loads values form environment variables
+          # e.g. os.getenv('Password'), os.environ["password"], env("Password")
```
loris-s-sonarsource, line 34: I suggest removing all these lines to provide examples, as well as the why of patternNot. Such as :

``` yaml
          # Potential FPs here are environment variables, templates and string substitutions, such as
          # - $my_password
          # - [My Password]
          # - <My Password>
          # - %s
          # - {{MyPassword}}
          # - $(mypassword)
          # - ${mypassword}
          # - getenv("PASS")
```

To do so, just use chatgpt with this prompt:

```
generate a string that would satisfy a given regex without telling anything else. say YES if you understand the assignment.
```

and then send the big regex with markdown backticks, it explains everything

Then, ensure that everything is actually right and add them as comments. I think this way it's gonna be very clear and great for maintainability 👍

loris-s-sonarsource (changes requested):
Please switch the comments to what I suggested. It's gonna look great 👍 (I hope)

Then, please add this case, reported by a user:

``` cpp
swprintf(connectionString, 1024, L"DRIVER=%ls;ServerName=%ls;Port=%ls;UID=%ls;Database=%ls;PWD=%ls;SSLMode=%ls",...);
```

For the environment let's try to be a little narrower, for example by checking only for 

- process.env.VAR for JS
- os.getenv for python
- etc. (see the excel file)

daniel-teuchert-sonarsource:
> For the environment let's try to be a little narrower, for example by checking only for
> 
> * process.env.VAR for JS
> * os.getenv for python
> * etc. (see the excel file)

I tried to narrow it down with the "\b". What kind of false negative do you think could happen with this?

loris-s-sonarsource (approved):
LGTM. I tried experimenting and creating a provider post block but it looks like it is not really possible

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/postgresql.yaml
```diff
@@ -27,10 +27,27 @@ provider:
...
+          patternNot:
+            "(?i)\
+            ^(\\$[a-z_]*)?(db|my)?_?pass(word|wd)?|\
+            ^\\[[\\w\\t \\-]+\\]$|\
+            ^<[\\w\\t \\-]+>$|\
+            ^%s$|\
+            ^\\{+[^}]*\\}+$|\
+            ^\\$\\(\\w+\\)$|\
+            \\$\\{\\w+(:-\\w+)?\\}|\
+            \\b(get)?env(iron)?\\b"
```
loris-s-sonarsource, line 50: I suggest removing as much patternNot as possible to try to clean up as much as we can, and adding that to the provider detection block instead:

```
  detection:
    post:
      # Potential FPs here are environment variables, templates and string substitutions, such as
      # - $my_password
      # - $db_password
      # - [mypass123]
      # - <example-password>
      # - %s
      # - {{MyPassword}}
      # - $(mypassword)
      # - ${mypassword}
      # - os.getenv("PASS")
      # - os.environ['PASS']
      # - `myPassword`
      # - *********
      # - None
      patternNot:
        "(?i)\
        ^(\\$[a-z_]*)?(db|my)?_?pass(word|wd)?|\
        ^\\[[\\w\\t \\-]+\\]$|\
        <[\\w\\t \\-]+>?|\
        ^%s$|\
        ^\\{+[^}]*\\}+$|\
        ^\\$\\(\\w+\\)$|\
        \\$\\{\\w+(:-\\w+)?\\}|\
        ^\\$[{(]|\
        \\b(get)?env(iron)?\\b|\
        ^`[a-z]+|\
        ^\\*{3,}$|\
        ^None$"
```

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/postgresql.yaml
```diff
@@ -164,6 +195,21 @@ provider:
...
+        - text: |
+            ENV PG_USER=os.getenv("PG_USER")
+            ENV PG_PASSWORD=os.getenv("PG_PASSWORD")
+            ENV PG_DATABASE=os.getenv("PG_DATABASE")
+          containsSecret: false
+        - text: |
+            ENV PG_USER=os.environ['PG_USER']
+            ENV PG_PASSWORD=os.environ['PG_PASSWORD']
+            ENV PG_DATABASE=os.environ['PG_DATABASE']
+          containsSecret: false
```
loris-s-sonarsource, line 212: I suggest removing the "ENV" since it's python and not Dockerfiles

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/postgresql.yaml
```diff
@@ -222,6 +276,21 @@ provider:
...
```
loris-s-sonarsource, line 293: same about dockerfile and python

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/postgresql.yaml
```diff
@@ -242,7 +311,11 @@ provider:
           pattern: "\\bPGPASSWORD=(?!\\\\?[\"'])([^\\s;]+).{1,40}\\bpsql\\b"
         post:
           # Avoid matching values found on SourceGraph that look like dummy passwords or insertions.
-          patternNot: "^(db|my)?_?pass(word|wd)?$|^\\*{3,}$|^\\$[A-Z_]+$|^\\$[{(]|^\\{{2,}"
+          patternNot: "^(db|my)?_?pass(word|wd)?$|\
```
loris-s-sonarsource, line 314:
```suggestion
          patternNot:
            "^(db|my)?_?pass(word|wd)?$|\            
```

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/postgresql.yaml
```diff
@@ -126,7 +146,18 @@ provider:
           pattern: "\\bPG_PASSWORD=(?!\\\\?[\"'])([^\\s]+)(?:$|\\s)"
```
loris-s-sonarsource, line 146:
```suggestion
          # looks for occurrences of the string `PG_PASSWORD=` followed by a value that is not enclosed in single or double quotes.
          pattern: "\\bPG_PASSWORD=(?!\\\\?[\"'])([^\\s]+)(?:$|\\s)"
```

loris-s-sonarsource (changes requested):
technically LGTM but maintainability requests

loris-s-sonarsource (approved):
LGTM !

PR #209, opened 2023-08-28
Modify recaptcha & Azure: Add FP and clarify

loris-s-sonarsource:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-recaptcha.yaml
```diff
@@ -20,7 +20,13 @@ provider:
         post:
           statisticalFilter:
             threshold: 3.9 # Based on significant sampling, 4.0 < entropy < 4.9
```
egon-okerman-sonarsource, line 22: Small question as I came across this PR: why is entropy 3.9 if the comment says "4.0 < entropy < 4.9"?

daniel-teuchert-sonarsource (approved):
LGTM!
I also have the same question as Egon.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-recaptcha.yaml
loris-s-sonarsource: Good point. I think we need Gaetan for that. I am going to keep it as is, just in case, and link this message to him when he gets back from holiday as I don't know if the comment has a typo or if the threshold itself has a problem

nils-werner-sonarsource: PR is merged, but the question seems still to be unresolved. Any follow-up here?

loris-s-sonarsource: Yes Nils, as written this will be discussed with Gaetan. This is in my action items.

PR #215, opened 2023-08-29
APPSEC-1007 Create S6732: Stripe API Keys should not be disclosed

loris-s-sonarsource:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/stripe.yaml
```diff
@@ -2,13 +2,13 @@ provider:
   metadata:
     name: Stripe Webhook Secret
```
daniel-teuchert-sonarsource, line 3: We probably should also change the name

daniel-teuchert-sonarsource (approved):
LGTM!
Only html files missing now.

PR #218, opened 2023-08-30
APPSEC-1039: S6736(secrets) Detect AMQP URLs

daniel-teuchert-sonarsource:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/rabbitmq.yaml
```diff
@@ -0,0 +1,80 @@
...
+          # - amqp://foo:bar@127.0.0.1:5672/"
+          # - amqp://guest:**@healthy.example.com
+          # - amqp://admin:#{ENV.fetch("ADMIN_PASSWORD")}@mqtt:5672
+          # - amqp://admin:%s@host:5672
+          # - amqp://admin:$env{admin_pwd}@host
+          patternNot: 
+            "(?i)\
+            (\\w)\\1{3,}|\
+            :guest|\
+            :password|\
```
loris-s-sonarsource, line 30: Honestly I don't think we should filter this out

loris-s-sonarsource (commented):
LGTM!

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/rabbitmq.yaml
```diff
@@ -0,0 +1,69 @@
...
+          # Type 1 pattern amqp(s)?://{username}:{password}@{host}:{port}/{path}
+          pattern: "\\bamqp(?:s)?://(\\w+:[^@\\s]+)@[-\\w%.\\+]+\\b(?:[-\\w%.\\+()@:~#?&/=]*)"
+        post:
+          # Avoid matching values found on SourceGraph that look like dummy passwords or insertions like:
+          # - amqp://admin:{{rabbitmq_password}}@localhost:8080/endpoint
+          # - amqp://foo:bar@127.0.0.1:5672/"
+          # - amqp://guest:**@healthy.example.com
+          # - amqp://admin:#{ENV.fetch("ADMIN_PASSWORD")}@mqtt:5672
+          # - amqp://admin:%s@host:5672
+          # - amqp://admin:$env{admin_pwd}@host
```
loris-s-sonarsource, line 23:
```suggestion
          #patternNot:
          #  # amqp://foo:bar@127.0.0.1:5672/
          #  - "(?i)foo|bar"
          #  # amqp://admin:#{ENV.fetch("ADMIN_PASSWORD")}@mqtt:5672
          #  - "(?i)\\b(get)?env(iron)?\\b"
          #  - "123456"
          #  - "(\\w)\\1{3,}"
          # # amqp://guest:**@healthy.example.com
          #  - ":(\\*)\\1{1,}"
          #  # amqp://admin:{{rabbitmq_password}}@localhost:8080/endpoint
          #  - ":\\{\\{[^}]+}}"
          #  # amqp://admin:%s@host:5672
          #  - ":%s\\b"
```

daniel-teuchert-sonarsource: I don't know why you want to write it like `:(\\*)\\1{1,}`. The main semantical difference I see is that it also takes the colon into account. Otherwise it also just looks for at least two `*` (directly after a colon). I would write it like `:\\*{2,} ` now.
Did you want to match anything else as well?

PR #222, opened 2023-08-31
S6694 Improved Mongo DB secret detection precision

daniel-teuchert-sonarsource:
We received feedback via Sonarcloud about Mongo DB placeholder credentials that were detected by rule S6652. The placeholder Mongo DB looked like this:
mongodb://xxx:xxx@mongo:8080/

This PR was created to prevent such false positives in the future.

To prevent this, repeating characters are recognized by the patternNot and will prevent raising issues for passwords like:
- xxx
- \*\*\*
- ...


loris-s-sonarsource (approved):
LGTM !

PR #224, opened 2023-09-04
APPSEC-1061 Modify S6292(secrets): Remove refs and add tests

loris-s-sonarsource:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/mws.yaml
```diff
@@ -2,32 +2,40 @@ provider:
...
-        type: Standards
-      - description: SANS Top 25 - Porous Defenses
-        link: https://www.sans.org/top25-software-errors/#cat3
-        type: Standards
   detection:
+    pre:
+      include:
+        content:
+          - "amazon"
+          - "mws"
```
daniel-teuchert-sonarsource, line 10: Does it make sense to use this here? This is also part of the start of the regex.

daniel-teuchert-sonarsource (approved):
LGTM. Not sure though if "mws" needs to be added to the pre filter.

PR #227, opened 2023-09-05
APPSEC-1056 S6336(secrets): Improvements to raise less FPs

daniel-teuchert-sonarsource:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/alibaba.yaml
```diff
@@ -64,30 +69,29 @@ provider:
...
       detection:
+        pre:
+          include:
+            content:
+              - aliyun
+              - alibaba
+              - alicloud
         matching:
-          pattern: "(?i)(?<![A-Z])ali(?:yun|baba|cloud).{0,50}['\"`]([0-9a-z]{30})['\"`]"
+          pattern: "(?i)(?:SECRET_?(?:ACCESS)?_?KEY|(?:ACCESS)?_?KEY_?SECRET)\\b[^0-9a-z]{0,10}([0-9a-z]{30})(?![a-z0-9\\/+=$\\-_])"
```
loris-s-sonarsource, line 83: What is the why for this change ? you removed pretty important prefixes here

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/alibaba.yaml
```diff
@@ -64,30 +69,29 @@ provider:
...
           containsSecret: true
           match: l0GdwcDYdJwB1VJ5pv0ormyTV9nhvW
-      detection:
-        matching:
-          pattern: "(?i)(?:SECRET_?(?:ACCESS)?_?KEY|(?:ACCESS)?_?KEY_?SECRET)\\b[^0-9a-z]{0,10}([0-9a-z]{30})(?![a-z0-9\\/+=$\\-_])"
-        post:
-          statisticalFilter:
-            threshold: 4.2
-          patternNot:
-            - "EXAMPLE"
```
loris-s-sonarsource, line 93: I think you mixed up the detection patterns between two different detection logics

loris-s-sonarsource (changes requested):

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/alibaba.yaml
daniel-teuchert-sonarsource: I am pretty sure that this is only a confusing diff display.
Basically we have one pattern with the ali.. prefix and one with the SECRET... prefix.
I added pre filters to both to decrease the false positives.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/alibaba.yaml
daniel-teuchert-sonarsource: No, this is only a weird diff display. The structure of the file changed a bit. The actual changes to the regex are minor.

loris-s-sonarsource: Ah ok I see

loris-s-sonarsource (approved):
LGTM ! That diff's really weird

PR #229, opened 2023-09-07
APPSEC-1073 Create S6751(secrets): Add PyPI 

egon-okerman-sonarsource:


gaetan-ferry-sonarsource (changes requested):
LGTM with a slight improvement suggestion.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/pypi.yaml
```diff
@@ -0,0 +1,44 @@
+provider:
+  metadata:
+    name: PyPI API tokens should not be disclosed
```
gaetan-ferry-sonarsource, line 3: I think you got confused with this change. This is supposed to contain the name of the provider (which it previously did).

gaetan-ferry-sonarsource (approved):
There is a small thing to fix, other than that it LGTM. I still approve the PR to avoid another back and forth.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/pypi.yaml
egon-okerman-sonarsource: My bad. What is `name` used for in the rule? Because I've seen it contain both the [secret](https://github.com/SonarSource/sonar-text/blob/master/sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/alchemy.yaml#L11) and [the name of the rule](https://github.com/SonarSource/sonar-text/blob/master/sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/airtable.yaml#L16) interchangeably (not in `provider` though, so I'll change this ASAP).

gaetan-ferry-sonarsource: In my mind, until now, it was containing the name of the rule. That said, I don't think this is a big deal, this information is not used afterward to the best of my knowledge.

PR #230, opened 2023-09-08
SONARTEXT-76 Improve messages for failing tests

jonas-wielage-sonarsource:


sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/utils/AbstractRuleExampleTest.java
```diff
@@ -112,6 +113,53 @@ private String displayName(Rule rule, RuleExample example) {
...
+    }
+
+    sb.append("\" failed as it matched on the following content:\n");
+    for (Issue issue : issues) {
+      sb.append("- \"");
+      try {
+        sb.append(retrieveMatchedContent(issue, ruleExample.getText()));
+      } catch (NullPointerException | IndexOutOfBoundsException e) {
+        sb.append("--- Could not calculate matched content cause of ");
+        sb.append(e.getClass().getSimpleName());
```
petertrr, line 132: Maybe log the stacktrace here too? Like `logger.error("Exception calculating matched content, see test results below", e)`?

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/utils/AbstractRuleExampleTest.java
```diff
@@ -112,6 +113,53 @@ private String displayName(Rule rule, RuleExample example) {
...
+      }
+      sb.append("\"\n");
+    }
+
+    if (ruleExample.isContainsSecret()) {
+      sb.append("But the example should have matched on: \"");
+      sb.append(ruleExample.getMatch());
+      sb.append("\"");
+    }
```
petertrr, line 141:
```suggestion
    } else {
      sb.append("But example doesn't expect a match");
    }
```

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/utils/AbstractRuleExampleTest.java
```diff
@@ -112,6 +113,53 @@ private String displayName(Rule rule, RuleExample example) {
...
+  private String failingMessage(Collection<Issue> issues, Rule rule, RuleExample ruleExample) {
+    StringBuilder sb = new StringBuilder();
+    sb.append("Test case \"");
+    sb.append(displayName(rule, ruleExample));
+    if (issues.isEmpty()) {
+      sb.append("\" failed as it didn't match on any content");
+      return sb.toString();
+    }
+
+    sb.append("\" failed as it matched on the following content:\n");
```
petertrr, line 125: Should we care about using `System.lineSeparator()` here instead of `\n`? Or do IDEs display it correctly also on Windows?

petertrr (approved):
LGTM, nice improvement. Just a couple of suggestions below

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/utils/AbstractRuleExampleTest.java
jonas-wielage-sonarsource: Good catch, I will change it. It works on windows, I'm not sure about mac

PR #231, opened 2023-09-08
SONARTEXT-81 Fail code generation process if there are deserialization errors

jonas-wielage-sonarsource:


sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/utils/UpdatingSpecificationFilesGenerator.java
```diff
@@ -59,11 +66,17 @@ class UpdatingSpecificationFilesGenerator {
...
 
+  @RegisterExtension
+  LogTesterJUnit5 logTester = new LogTesterJUnit5();
+
   // Suppress warning, as there are no assertions inside here
   @Test
   @SuppressWarnings("java:S2699")
   void firstStep() {
     writeSpecificationFileDefinition();
+    assertThatNoException().isThrownBy(SpecificationLoader::new);
```
petertrr, line 77: Won't this always succeed as we capture deserialization exceptions inside `SpecificationLoader`?

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/utils/UpdatingSpecificationFilesGenerator.java
```diff
@@ -59,11 +66,17 @@ class UpdatingSpecificationFilesGenerator {
...
+    List<LogAndArguments> errorLogs = logTester.getLogs(Level.ERROR);
+    assertThat(errorLogs).withFailMessage(() -> failMessage(errorLogs)).isEmpty();
```
petertrr, line 79: I would suggest to avoid test assertions here. As it is only a workaround to put generation logic in a test, I think it may cause problems if we rely on AssertJ here if (or when) we move this codegen elsewhere. Simply checking condition and throwing a RuntimeException with a message would be enough IMO

sonar-text-plugin/src/test/java/org/sonar/plugins/secrets/utils/UpdatingSpecificationFilesGenerator.java
jonas-wielage-sonarsource: Yes this assert can be removed

petertrr (approved):
LGTM!

PR #235, opened 2023-09-12
APPSEC-1068 Create S6760 (secrets): Add Yandex tokens

sebastien-andrivet-sonarsource:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/yandex.yaml
```diff
@@ -0,0 +1,50 @@
...
+      metadata:
+        name: Yandex IAM tokens should not be disclosed
+        message: Make sure this Yandex IAM token gets revoked, changed, and removed from the code.
+      detection:
+        pre:
+          include:
+            content:
+              - t1.
+        matching:
+          pattern: "\\b(t1\\.[\\w-]+[=]{0,2}\\.[\\w-]{86}[=]{0,2})\\b"
```
egon-okerman-sonarsource, line 42: To limit FPs, I would put a minimal amount of characters on the first `[\\w-]+`. From the example 70 or 80 might be safe enough, but you know better what this threshold should be.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/yandex.yaml
```diff
@@ -0,0 +1,50 @@
...
+    name: Yandex
+    category: Cloud provider
+
+  rules:
+    - id: yandex_oauth_token
+      rspecKey: S6760
+      metadata:
+        name: Yandex OAuth tokens should not be disclosed
+        message: Make sure this Yandex OAuth token gets revoked, changed, and removed from the code.
+      detection:
```
egon-okerman-sonarsource, line 12: Detection can be improved by adding some post checks for dummy words like "test" or "example" (the CONTRIBUTING.md draft has a list of checks that are useful.)

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/yandex.yaml
egon-okerman-sonarsource, line 42: The end of the regex does not work correctly for secrets that end with `=`.

For example, for `t1.P=.2CVgkS3oFhYzUA4NYeOd8VEIy8mBEY2q28T2Ro2pLyWiofIHRKJWqzahcKLwFS63jWhGq3rLwlu6oFsVswvrx9=` the match will be `t1.P=.2CVgkS3oFhYzUA4NYeOd8VEIy8mBEY2q28T2Ro2pLyWiofIHRKJWqzahcKLwFS63jWhGq3rLwlu6oFsVswvrx9`.

egon-okerman-sonarsource (changes requested):
Some changes are necessary still

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/yandex.yaml
sebastien-andrivet-sonarsource: The first part can be as small as 1 character.

egon-okerman-sonarsource: Oh okay, nevermind this then.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/yandex.yaml
sebastien-andrivet-sonarsource: I have added `patternNot` at the level of the provider and some tests for those cases.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/yandex.yaml
sebastien-andrivet-sonarsource: Fixed and I added a test for this case.

egon-okerman-sonarsource (approved):
LGTM

PR #237, opened 2023-09-12
Modify S6694: Add new FP detection

loris-s-sonarsource:
False positive on strings like `<:sonarloris:>`


daniel-teuchert-sonarsource (approved):
LGTM!

PR #249, opened 2023-09-13
Modify S6334: Add rspec checks and tweak message

loris-s-sonarsource:
The goal of this pr is to add checks and also make the message less strict, since this API key should in lot of cases be user-facing.


daniel-teuchert-sonarsource (approved):
LGTM!

PR #251, opened 2023-09-14
Modify S6290: Validation

loris-s-sonarsource:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/aws.yaml
```diff
@@ -2,46 +2,42 @@ provider:
...
+        post:
+          statisticalFilter: 
+            threshold: 4.2
+          patternNot:
+            # Character repeated 5 times
+            - "([\\w])\\1{5,}"
+            # Common text placeholders
+            - "(?i)(?:s|ex)ample|foo|bar|test|abcd|redacted|key"
+            # Common numeric placeholders
+            - "1234(?:56)"
```
daniel-teuchert-sonarsource, line 30: I'd say there is no need to check for any potential "56" if we already found 1234 since just 1234 will also match 123456.
```suggestion
            - "1234"
```

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/aws.yaml
```diff
@@ -59,20 +55,25 @@ provider:
...
```
daniel-teuchert-sonarsource, line 76: Same as above.
```suggestion
            - "1234"
```

daniel-teuchert-sonarsource (approved):
LGTM!

PR #252, opened 2023-09-14
Create S6773: Hashicorp tokens should not be disclosed(APPSEC-1066)

sebastien-andrivet-sonarsource:


gaetan-ferry-sonarsource (changes requested):
LGTM with a slight adjustement suggestion.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/hashicorp.yaml
```diff
@@ -0,0 +1,32 @@
...
+      metadata:
+        name: Hashicorp Batch tokens should not be disclosed
+      detection:
+        pre:
+          include:
+            content:
+              - b.AAAAAQ
+              - hvb.AAAAAQ
+        matching:
+          pattern: "\\b((?:hv)?b\\.AAAAAQ[\\w-]{120,240})\\b"
```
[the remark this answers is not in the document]
sebastien-andrivet-sonarsource: Agree, the `\b` at the end is not adequate. In addition, from what I can see, there is no actual maximum. So maybe, it would be better to remove the maximum. This is what I have done in the update regex,

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/hashicorp.yaml
```diff
@@ -0,0 +1,32 @@
...
+      metadata:
+        name: Hashicorp Batch tokens should not be disclosed
+      detection:
+        pre:
+          include:
+            content:
+              - b.AAAAAQ
+              - hvb.AAAAAQ
+        matching:
+          pattern: "\\b((?:hv)?b\\.AAAAAQ[\\w-]{120,})"
```
gaetan-ferry-sonarsource, line 19: I think this is exactly the same as `\b((?:hv)?b\.AAAAAQ[\w-]{120})`, at least from a detection point of view. You should consider removing the `,` to avoid possible backtracking issues.

gaetan-ferry-sonarsource (approved):
LGTM. I propose an optimization that you can accept or not.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/hashicorp.yaml
sebastien-andrivet-sonarsource: Doing so will be less user-friendly (only a part of the token will be highlighted) so I keep it as it is.

PR #254, opened 2023-09-14
Modify S6335: Remove FPs

loris-s-sonarsource:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/gcp.yaml
```diff
@@ -2,43 +2,33 @@ provider:
...
     - rspecKey: S6335
       id: google-service-account-keys
       metadata:
         name: Google Service Account Keys
+      detection:
+        # Matches multiline private keys or single line private keys containing '\n's
+        matching:
+          pattern: "(\
+                    -----BEGIN PRIVATE KEY-----\
+                    (?:\\n|\\\\{1,2}n)\
```
daniel-teuchert-sonarsource, line 27: Maybe match also other whitespaces here? The key could be indented. You could also first match for the line break and only afterwards accept other whitespaces.
```suggestion
                    (?:\\s|\\\\{1,2}n)++\
```

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/gcp.yaml
```diff
@@ -2,43 +2,33 @@ provider:
...
+                    (?:[0-9A-Za-z/+]{80,}|(?:[0-9A-Za-z/+]{40,}(?:\\n|\\\\{1,2}n))+[0-9A-Za-z/+=]+)={0,2}\
+                    (?:\\n|\\\\{1,2}n)\
```
daniel-teuchert-sonarsource, line 29: Same as above about potential indentation.

sonar-text-plugin/src/test/java/org/sonar/plugins/common/TextAndSecretsSensorTest.java
```diff
@@ -73,20 +73,6 @@ void describe() {
     assertThat(logTester.logs()).isEmpty();
   }
 
-  @Test
```
daniel-teuchert-sonarsource, line 76: Why is this removed? Was it removed automatically when running the script to create the tests?

daniel-teuchert-sonarsource (changes requested):
Overall good. Minor request about indentation.
Also if you want you can think about adjusting the regex to also find stuff like:
```
private_key = "-----BEGIN PRIVATE KEY----- " \
            + "asdjasdjkasdnasd..." \
            + "-----END PRIVATE KEY-----";
```
Of course, this might look differently in different languages so I am not sure if this would be too much effort and not the goal of this PR right now.

sonar-text-plugin/src/test/java/org/sonar/plugins/common/TextAndSecretsSensorTest.java
loris-s-sonarsource: No, I removed it because it is an artifact from the beginnings of sonar-text

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/gcp.yaml
loris-s-sonarsource: done !

daniel-teuchert-sonarsource (approved):
LGTM!
I would suggest to rename the test file to GCPCheckTest.java since this is probably the name our tool would give it automatically. I think it is usually {name_of_the_provider}CheckTest.java

PR #258, opened 2023-09-15
Modify Azure checks: Add pre checks for efficiency

loris-s-sonarsource:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/azure.yaml
```diff
@@ -199,7 +212,12 @@ provider:
       rspecKey: S6684
       metadata:
         name: Azure subscription keys should not be leaked
-      detection:
+         pre:
+          include:
+            content:
```
egon-okerman-sonarsource, line 217: Add "microsoft.com" as well?

loris-s-sonarsource: yup, done !

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/azure.yaml
```diff
@@ -200,6 +213,12 @@ provider:
       metadata:
         name: Azure subscription keys should not be leaked
       detection:
+        pre:
+          include:
+            content:
+              - "secret"
+              - "key"
+              - "microsoft\\.com"
```
loris-s-sonarsource, line 221:
```suggestion
              - "microsoft.com"
```

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/azure.yaml
```diff
@@ -107,6 +107,10 @@ provider:
       metadata:
         name: Azure Storage Account Keys
       detection:
+        pre:
+          include:
+            content:
+              - "AccountKey"
```
egon-okerman-sonarsource, line 113: This can even be `AccountKey=` to be more specific.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/azure.yaml
```diff
@@ -200,6 +213,12 @@ provider:
       metadata:
         name: Azure subscription keys should not be leaked
       detection:
+        pre:
+          include:
+            content:
```
egon-okerman-sonarsource, line 218: The `content` pre-check is an AND list, correct? If so, we might miss some things by filtering on both "secret" and "key", if the pattern only requires one of them.

egon-okerman-sonarsource (commented):
I have some questions first, before I can approve this.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/azure.yaml
loris-s-sonarsource: No, I think that's an OR, but I'm gonna ask Jonas

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/azure.yaml
loris-s-sonarsource: very true

```suggestion
              - "AccountKey="
```

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/azure.yaml
loris-s-sonarsource: it's an OR, he confirmed

egon-okerman-sonarsource: In that case, I would not filter on "secret" or "key" since they will be more generic than "microsoft.com"

loris-s-sonarsource: totally yeah actually

egon-okerman-sonarsource (approved):
LGTM

PR #259, opened 2023-09-15
Create S6777: Shippo tokens should not be disclosed(APPSEC-1070)

sebastien-andrivet-sonarsource:


egon-okerman-sonarsource (approved):
LGTM

PR #260, opened 2023-09-15
Modify Google Oauth: Made the regexes easier to read

loris-s-sonarsource:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-oauth2.yaml
```diff
@@ -55,16 +55,36 @@ provider:
...
+          #
+          #   The further that the target secret is from the keyword "secret",
+          #   the less likely it is to be related.
+          #
+          # 3. Then looks for the target secret.
+          pattern: "(?i)\
+                    (?:\
+                        \\.apps\\.googleusercontent\\.com\\b\
+                        (?:[^})\\r\\n]*+\\r?\\n){0,5}\
+                        [^})\\r\\n]*\
```
loris-s-sonarsource, line 81: This line makes this regex vulnerable to polynomial runtime, I don't know how to proceed

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-oauth2.yaml
```diff
@@ -159,7 +179,23 @@ provider:
...
+          pattern: "(?i)\
+                    (?:\
+                        secret\
+                        (?:[^})\\r\\n]*+\\r?\\n)?\
+                        [^})\\r\\n]{0,50}\
+                    )\
+                    \\b([\\w\\-]{24})\\b\
+                    (?:\
+                        (?:[^})\\r\\n]*+\\r?\\n){0,3}\
+                        [^})\\r\\n]*\
```
loris-s-sonarsource, line 191: same here

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-oauth2.yaml
```diff
@@ -225,10 +261,17 @@ provider:
...
-          # second parameter (client secret) is a quoted string
-          pattern: "new\\s+google\\.auth\\.OAuth2\\(\\s*[\"'][a-z0-9\\-]+\\.apps\\.googleusercontent\\.com[\"']\\s*,\\s*+(?://[^\r\n]+\\s*+)?[\"']([A-Za-z0-9_\\-]{24})[\"']"
+          # - first parameter (client ID) is a quoted string ending in ".apps.googleusercontent.com"
+          # - optional "//" comment after the client ID
+          # - second parameter (client secret) is a quoted string
+          pattern: "new\\s++google\\.auth\\.OAuth2\\(\\s*+\
+                    [\"']\
+                        [a-z0-9\\-]++\\.apps\\.googleusercontent\\.com\
+                    [\"']\
+                    \\s*+,\\s*+(?://[^\r\n]++\\s*+)?\
```
daniel-teuchert-sonarsource, line 271: It seems to work, but we usually "double escape" \n and \r.
So I suggest either adding the additional backslash here for consistency or removing the additional backslash from the other instances.
```suggestion
                    \\s*+,\\s*+(?://[^\\r\\n]++\\s*+)?\
```

daniel-teuchert-sonarsource (changes requested):
Added some suggestions

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-oauth2.yaml
```diff
@@ -55,16 +55,36 @@ provider:
...
+          #    the start of the line (following line).
+          #   - 50 characters is an arbitrary distance and should allow for
+          #   whitespace and a prefix of some sort.
+          #
+          #   The further that the target secret is from the keyword "secret",
+          #   the less likely it is to be related.
+          #
+          # 3. Then looks for the target secret.
+          pattern: "(?i)\
+                    (?:\
```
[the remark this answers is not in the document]
daniel-teuchert-sonarsource: Maybe we do not need such long regexes if we use pattern-before or context. I suggest checking out the detections defined in ssh.yaml

daniel-teuchert-sonarsource (changes requested):

loris-s-sonarsource (commented):
Reworked and gave up on refactoring that regex

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-oauth2.yaml
```diff
@@ -55,16 +55,37 @@ provider:
...
+                        \\.apps\\.googleusercontent\\.com\\b\
+                        (?:[^})\\r\\n]*+\\r?\\n){0,5}\
+                        [^})\\r\\n]{0,600}\
```
daniel-teuchert-sonarsource, line 82: The 600 (5 * 120) does not really make sense here because you are not matching any line breaks.
So basically you are matching **one** line with up to 600 characters. I don't think this is your intention.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-oauth2.yaml
```diff
@@ -157,9 +178,25 @@ provider:
...
+          pattern: "(?i)\
+                    (?:\
+                        secret\
+                        (?:[^})\\r\\n]*+\\r?\\n)?\
+                        [^})\\r\\n]{0,50}\
+                    )\
+                    \\b([\\w\\-]{24})\\b\
+                    (?:\
+                        (?:[^})\\r\\n]*+\\r?\\n){0,3}\
+                        [^})\\r\\n]{0,360}\
```
daniel-teuchert-sonarsource, line 192: Same as with the 600 above.

daniel-teuchert-sonarsource (changes requested):
Sorry, there are still some things that look unintentional to me.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-oauth2.yaml
loris-s-sonarsource: Line breaks are represented by `\n` so I don't get what you mean: it's actually matching line breaks

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-oauth2.yaml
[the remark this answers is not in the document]
loris-s-sonarsource: I tried removing them and replacing them with backslashes and it breaks.. for some reason.. I don't understand

daniel-teuchert-sonarsource (approved):
I think this is as good as it gets for now. The rest will be handled via hardening.

PR #264, opened 2023-09-19
APPSEC-1129 Create S6782(secrets): Docker Hub tokens should not be disclosed

egon-okerman-sonarsource:


sebastien-andrivet-sonarsource (approved):
LGTM

PR #269, opened 2023-09-21
APPSEC-1125: S6688, S6692, S6700 Add examples from RSPEC

daniel-teuchert-sonarsource:


sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-recaptcha.yaml
```diff
@@ -27,6 +27,14 @@ provider:
...
             - "6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe"
       examples:
+        - text: |
+            # noncompliant example
+            props.set("recaptcha_secret", "6LcaQa4mAAAAAFvhmzAd2hErGBSt4FC-BPzm4cNS")
+          containsSecret: true 
+          match: 6LcaQa4mAAAAAFvhmzAd2hErGBSt4FC-BPzm4cNS
+        - text: |
+            # compliant example
+            props.set("recaptcha_secret", System.getenv("RECAPTCHA_SECRET"))
```
loris-s-sonarsource, line 37: It does not compile and I think it's because it's missing the match an containsSecret fields here.

Always try compiling before assigning for review ! 😄

loris-s-sonarsource (changes requested):

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-recaptcha.yaml
daniel-teuchert-sonarsource: Ohh, this is interesting.
I actually ran the test GoogleUniqueNameCheckTest and it was successful.
But when I run `mvn test` it also fails.

loris-s-sonarsource: Strange, I hope there's no issue with the tests 😅

loris-s-sonarsource (approved):
LGTM!

PR #270, opened 2023-09-21
Modify mysql rules: Remove FPs and improve syntax

loris-s-sonarsource:


daniel-teuchert-sonarsource (approved):
LGTM!
To go the extra mile you could add test cases for the newly filtered out FPs.

PR #285, opened 2023-09-26
SONARTEXT-88 Integrate rule-api via exec-maven-plugin

petertrr:


pom.xml
```diff
@@ -254,6 +257,70 @@
         <module>its</module>
       </modules>
     </profile>
+    <profile>
+      <id>enable-rule-api</id>
```
petertrr, line 261: Witohut a profile, something conflicts with RE scripts and the build fails somewhere in the beginning

README.md
```diff
@@ -54,20 +54,15 @@ rm empty-settings.xml
...
 ```
 
 ### Generate files to include new secrets
 
 After the change, addition or removal of secret specifications, this script can be run to generate the Java classes that are needed 
 for the inclusion or deletion of these secrets.
 
-To run this script you should specify the file name of the rule-api jar `<ruleApiFileName>`, otherwise a default value of `rule-api-snap.jar` will be assumed.
-The rule-api jar needs to be located in the root folder of the project.
-
```
jonas-wielage-sonarsource, line 70: Please include, that this script also takes care of generating static rspec files for new rules automatically.

pom.xml
```diff
@@ -73,6 +73,7 @@
     <sonar.api.version>10.1.0.809</sonar.api.version>
     <analyzer.commons.version>2.7.0.1482</analyzer.commons.version>
     <orchestrator.version>3.40.0.183</orchestrator.version>
+    <rules-api.version>2.6.0.2454</rules-api.version>
```
jonas-wielage-sonarsource, line 76: We have a newer version published. `2.7.0.2612`

jonas-wielage-sonarsource (approved):
LGTM, please see my suggestions.
Also, please test both generate/update and the changed script locally beforehand if not done yet.

PR #287, opened 2023-09-27
Update README.md

agigleux:
It was weird to have a link mentioning 110+ while clicking you reach a web site with "only" 60 rules. This is expected and like this by design because one rule can detect multiple secret patterns.


jonas-wielage-sonarsource (approved):
LGTM

PR #288, opened 2023-09-28
SONARTEXT-77 Record and log time spent on different analysis stages

petertrr:


sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SpecificationBasedCheck.java
```diff
@@ -37,7 +40,8 @@ public abstract class SpecificationBasedCheck extends Check {
   private List<SecretMatcher> matcher;
 
   // shared map between all specificationBasedChecks used to calculate if secret was already reported for textRange
-  private Map<InputFileContext, List<TextRange>> reportedIssuesForCtx;
+  private Map<URI, List<TextRange>> reportedIssuesForCtx;
```
petertrr, line 43: SQ complained that `InputFileContext` is not `Comparable`, so I used the thing that plugin-api claims to be unique among the files

sonar-text-plugin/src/test/java/org/sonar/plugins/common/DurationStatisticsTest.java
```diff
@@ -0,0 +1,55 @@
...
+      return null;
+    });
+    durationStatistics.log();
+
+    assertThat(durationStatistics.stats)
+      .hasSize(1);
+    assertThat(logTester.logs())
+      .anyMatch(s -> s.startsWith("Duration Statistics"))
+      .anyMatch(s -> s.contains("1 times"));
+  }
```
mstachniuk, line 54: I would like to see more tests:
- when one check is executed a few times (to verify the mean number) 
- when more checks are executed
I know that the output with be not deterministic because of `Thread.sleep(100)`, but I would like to see the whole format of the output.

sonar-text-plugin/src/test/java/org/sonar/plugins/common/DurationStatisticsTest.java
```diff
@@ -0,0 +1,55 @@
...
+        Thread.sleep(100);
+      } catch (InterruptedException e) {
+        throw new RuntimeException(e);
+      }
+      return null;
+    });
+    durationStatistics.log();
+
+    assertThat(durationStatistics.stats)
+      .hasSize(1);
```
mstachniuk, line 50: Maybe this should be validated by `assertThat(logTester.logs(Level.INFO)).hasSize(1);`?
Then `stats` can be private.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/DurationStatistics.java
```diff
@@ -0,0 +1,118 @@
...
+
+public class DurationStatistics {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DurationStatistics.class);
+
+  private static final String PROPERTY_KEY = "sonar.text.duration.statistics";
+
+  // modifier is package-private to make visible in tests
+  final Map<String, Measurement> stats = new ConcurrentHashMap<>();
+
+  private final AtomicBoolean isRecordingEnabled = new AtomicBoolean(false);
```
mstachniuk, line 46: Why does this need to be atomic?

sonar-text-plugin/src/main/java/org/sonar/plugins/common/DurationStatistics.java
```diff
@@ -0,0 +1,118 @@
...
+import org.slf4j.LoggerFactory;
+import org.sonar.api.config.Configuration;
+
+public class DurationStatistics {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DurationStatistics.class);
+
+  private static final String PROPERTY_KEY = "sonar.text.duration.statistics";
+
+  // modifier is package-private to make visible in tests
+  final Map<String, Measurement> stats = new ConcurrentHashMap<>();
```
mstachniuk, line 44: I'm wonder why this class is not a copy from sonar-iac?

sonar-text-plugin/src/main/java/org/sonar/plugins/common/DurationStatistics.java
```diff
@@ -0,0 +1,118 @@
...
+      out.append("Duration Statistics");
+      out.append(formatEntries(format, stats.entrySet().stream().filter(s -> s.getKey().endsWith("-total"))));
+      var generalReport = out.toString();
+      LOGGER.info(generalReport);
+
+      out = new StringBuilder();
+      out.append("Granular Duration Statistics");
+      out.append(formatEntries(format, stats.entrySet().stream().filter(s -> !s.getKey().endsWith("-total"))));
+      var verboseReport = out.toString();
+      LOGGER.info(verboseReport);
```
mstachniuk, line 82: I guess here should be `.debug()`

sonar-text-plugin/src/main/java/org/sonar/plugins/common/DurationStatistics.java
```diff
@@ -0,0 +1,118 @@
...
+
+  public void log() {
+    if (isRecordingEnabled.get()) {
+      var out = new StringBuilder();
+      var symbols = new DecimalFormatSymbols(Locale.ROOT);
+      symbols.setGroupingSeparator('\'');
+      NumberFormat format = new DecimalFormat("#,###", symbols);
+      out.append("Duration Statistics");
+      out.append(formatEntries(format, stats.entrySet().stream().filter(s -> s.getKey().endsWith("-total"))));
+      var generalReport = out.toString();
```
mstachniuk, line 75: `generalReport` can be inlined

sonar-text-plugin/src/main/java/org/sonar/plugins/common/DurationStatistics.java
```diff
@@ -0,0 +1,118 @@
...
+    stats.computeIfAbsent(id, key -> new Measurement()).add(elapsedTime);
+  }
+
+  public void log() {
+    if (isRecordingEnabled.get()) {
+      var out = new StringBuilder();
+      var symbols = new DecimalFormatSymbols(Locale.ROOT);
+      symbols.setGroupingSeparator('\'');
+      NumberFormat format = new DecimalFormat("#,###", symbols);
+      out.append("Duration Statistics");
```
mstachniuk, line 73: The ticket says about printing N's slowest checks, but for me, it is better to print everything here

mstachniuk (changes requested):
Please see my comments

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SecretMatcher.java
```diff
@@ -33,34 +34,55 @@ public class SecretMatcher {
...
+    AuxiliaryPatternMatcher auxiliaryPatternMatcher,
+    Predicate<InputFileContext> preFilter,
+    Predicate<String> postFilter,
+    DurationStatistics ds) {
     this.rule = rule;
     this.patternMatcher = patternMatcher;
     this.auxiliaryPatternMatcher = auxiliaryPatternMatcher;
     this.preFilter = preFilter;
     this.postFilter = postFilter;
+    this.ds = ds;
```
jonas-wielage-sonarsource, line 50: Please change this variable name

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SecretMatcher.java
```diff
@@ -33,34 +34,55 @@ public class SecretMatcher {
...
     Predicate<String> postFilter = PostFilterFactory.createPredicate(rule.getDetection().getPost(), rule.getDetection().getMatching());
     AuxiliaryPatternMatcher auxiliaryMatcher = AuxiliaryPatternMatcherFactory.build(rule.getDetection().getMatching());
-    return new SecretMatcher(rule, patternMatcher, auxiliaryMatcher, preFilter, postFilter);
+    return new SecretMatcher(rule, patternMatcher, auxiliaryMatcher, preFilter, postFilter, durationStatistics);
   }
 
   public List<Match> findIn(InputFileContext fileContext) {
-    if (!preFilter.test(fileContext)) {
+    boolean isPreFilterMatching = ds.timed(getRuleId() + "-preFilter", () -> preFilter.test(fileContext));
+    if (!isPreFilterMatching) {
```
jonas-wielage-sonarsource, line 69:
```suggestion
    boolean rejectedOnPreFilter= ds.timed(getRuleId() + "-preFilter", () -> !preFilter.test(fileContext));
    if (rejectedOnPreFilter) {
```
Please don't name it matching as we have it enough already in our code base

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SpecificationBasedCheck.java
jonas-wielage-sonarsource: Fine for me

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SpecificationBasedCheck.java
```diff
@@ -48,29 +52,37 @@ protected SpecificationBasedCheck() {
...
     this.reportedIssuesForCtx = reportedIssuesForCtx;
+    this.durationStatistics = durationStatistics;
     String rule = ruleKey.rule();
     List<Rule> rulesForKey = loader.getRulesForKey(rule);
     if (rulesForKey.isEmpty()) {
       LOG.error("Found no rule specification for rule with key: {}", rule);
     }
     this.matcher = rulesForKey.stream()
-      .map(SecretMatcher::build)
+      .map(r -> SecretMatcher.build(r, durationStatistics))
```
jonas-wielage-sonarsource, line 70:
```suggestion
      .map(rule -> SecretMatcher.build(rule, durationStatistics))
```

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SpecificationBasedCheck.java
```diff
@@ -48,29 +52,37 @@ protected SpecificationBasedCheck() {
...
       .collect(Collectors.toList());
   }
 
   @Override
   public void analyze(InputFileContext ctx) {
     for (SecretMatcher secretMatcher : matcher) {
-      secretMatcher.findIn(ctx).stream()
+      durationStatistics.timed(secretMatcher.getRuleId() + "-total", () -> secretMatcher.findIn(ctx))
```
jonas-wielage-sonarsource, line 77: I would suggest not concatenating it with `-` but with `::` instead. This way it's more obvious where the rule Id ends and where the suffix starts.
Same for the other names.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/DurationStatistics.java
petertrr: Doesn't necessarily need to be, though having `final` field with mutable state somehow appeals to me more than having a mutable class field. OTOH, `stats` doesn't need to be `ConcurrentHashMap` either, but we had it like that in iac.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/DurationStatistics.java
petertrr: I iac we only gather total time for each `time`-wrapped invocation. Here, if, for example, a file is rejected by pre-filter, the main matcher is not executed, and this info could be useful for analysis. That's why I decided to keep both total time and total count.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/DurationStatistics.java
petertrr: Debug output of scanner would post to much, moreover we only intend to enable this reporting on our test projects. However, having them split into two sections improves readability

sonar-text-plugin/src/main/java/org/sonar/plugins/common/DurationStatistics.java
petertrr: I think there was some sort of finding by SQ, that's why we un-inlined it in the first place =D But I'll double check

sonar-text-plugin/src/main/java/org/sonar/plugins/common/DurationStatistics.java
petertrr: Agree, this is also related to why verbose report is not under DEBUG level

sonar-text-plugin/src/main/java/org/sonar/plugins/common/DurationStatistics.java
petertrr: Yes, SQ complains that `out.toString()` will be executed even if logging level is higher than INFO. But StringBuilder can be inlined without explicit call to `toString()`

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SpecificationBasedCheck.java
petertrr: Did some renaming around, because `ruleKey.rule()` is actually a `ruleId`

sonar-text-plugin/src/main/java/org/sonar/plugins/secrets/api/SecretMatcher.java
```diff
@@ -33,34 +34,61 @@ public class SecretMatcher {
...
     Predicate<InputFileContext> preFilter = PreFilterFactory.createPredicate(rule.getDetection().getPre());
     Predicate<String> postFilter = PostFilterFactory.createPredicate(rule.getDetection().getPost(), rule.getDetection().getMatching());
     AuxiliaryPatternMatcher auxiliaryMatcher = AuxiliaryPatternMatcherFactory.build(rule.getDetection().getMatching());
-    return new SecretMatcher(rule, patternMatcher, auxiliaryMatcher, preFilter, postFilter);
+    return new SecretMatcher(rule, patternMatcher, auxiliaryMatcher, preFilter, postFilter, durationStatistics);
   }
 
   public List<Match> findIn(InputFileContext fileContext) {
-    if (!preFilter.test(fileContext)) {
+    boolean isRejectedOnPreFilter = durationStatistics.timed(
```
mstachniuk, line 68: Consider providing methods like `timedPreFilter`, timedSuffixMatcher`, timedSuffixPost` etc. 
IMO will improve readability and will not leak `DurationStatistics.SUFFIX_PRE` and other constants to client code.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/DurationStatistics.java
```diff
@@ -0,0 +1,120 @@
...
+
+  void addRecord(String id, long elapsedTime) {
+    stats.computeIfAbsent(id, key -> new Measurement()).add(elapsedTime);
+  }
+
+  public void log() {
+    if (isRecordingEnabled.get()) {
+      var symbols = new DecimalFormatSymbols(Locale.ROOT);
+      symbols.setGroupingSeparator('\'');
+      NumberFormat format = new DecimalFormat("#,###", symbols);
```
mstachniuk, line 75: Maybe this should be moved to `formatEntry` or to class level as we always use the same formatting?

mstachniuk (approved):
LGTM

PR #289, opened 2023-09-28
APPSEC-1158: Add patternNot to GitHub secret detection (S6689)

daniel-teuchert-sonarsource:
This PR adds a patternNot filter to the GitHub secret detection to reduce false positives.


egon-okerman-sonarsource (approved):
LGTM

PR #299, opened 2023-10-04
SONARTEXT-96 Improve setup of integration tests in Gradle

mstachniuk:


build-logic/src/main/kotlin/org.sonarsource.text.integration-test.gradle.kts
```diff
@@ -0,0 +1,27 @@
...
+  testClassesDirs = integrationTest.output.classesDirs
+  classpath = configurations[integrationTest.runtimeClasspathConfigurationName] + integrationTest.output
+}
+
+tasks.check {
+  dependsOn(integrationTestTask)
+}
+
+dependencies {
+  "integrationTestImplementation"(project)
```
petertrr, line 26: I'm not sure if this is needed here as we are testing an artifact from another subproject

mstachniuk: Indeed, not needed.

PR #300, opened 2023-10-04
SONARTEXT-79 Add property for user-defined include patterns

rudy-regazzoni-sonarsource:


sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextAndSecretsSensor.java
```diff
@@ -42,24 +37,20 @@
...
-import org.sonarsource.analyzer.commons.ProgressReport;
 
 public class TextAndSecretsSensor implements Sensor {
 
-  private static final Logger LOG = LoggerFactory.getLogger(TextAndSecretsSensor.class);
-
   public static final String EXCLUDED_FILE_SUFFIXES_KEY = "sonar.text.excluded.file.suffixes";
-
   private static final String ANALYZE_ALL_FILES_KEY = "sonar.text.analyzeAllFiles";
+  public static final String INCLUDED_FILE_SUFFIXES_KEY = "sonar.text.include.file.suffixes";
```
jonas-wielage-sonarsource, line 45:
```suggestion
  public static final String INCLUDED_FILE_SUFFIXES_KEY = "sonar.text.included.file.suffixes";
```

sonar-text-plugin/src/main/java/org/sonar/plugins/common/Analyzer.java
```diff
@@ -0,0 +1,144 @@
...
+
+  public Analyzer notBinaryFilePredicate(NotBinaryFilePredicate notBinaryFilePredicate) {
+    this.notBinaryFilePredicate = notBinaryFilePredicate;
+    return this;
+  }
+
+  public Analyzer analyzeAllFilesMode(boolean analyzeAllFilesMode) {
+    this.analyzeAllFilesMode = analyzeAllFilesMode;
+    return this;
+  }
```
jonas-wielage-sonarsource, line 65: Here we should use a builder.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/Analyzer.java
```diff
@@ -0,0 +1,144 @@
...
+  private boolean displayHelpAboutExcludingBinaryFile = true;
+
+  private SensorContext sensorContext;
+  private List<Check> activeChecks;
+  private NotBinaryFilePredicate notBinaryFilePredicate;
+  private boolean analyzeAllFilesMode;
+
+  public Analyzer() {
+    // Empty analyzer by default, configuration provided by setters
+  }
```
jonas-wielage-sonarsource, line 45: Looks more like a builder pattern should be used.

jonas-wielage-sonarsource: But I'm not sure this is necessary. The analyzer mainly depends on `sensorContext` and the files. Why don't we give everything at once to a static method of the Analyzer without constructing the Analyzer?

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextAndSecretsSensor.java
```diff
@@ -76,42 +67,32 @@ public void describe(SensorDescriptor descriptor) {
...
 
-    NotBinaryFilePredicate notBinaryFilePredicate = binaryFilePredicate(sensorContext);
-    FilePredicate filePredicate = isSonarLintContext(sensorContext) || analyzeAllFiles(sensorContext)
-      ? notBinaryFilePredicate
-      : LANGUAGE_FILE_PREDICATE;
+    // Retrieve list of files to analyse using the right FilePredicate
+    boolean analyzeAllFiles = isSonarLintContext(sensorContext) || analyzeAllFiles(sensorContext);
+    var notBinaryFilePredicate = binaryFilePredicate(sensorContext);
+    var textFilePredicate = textFilePredicate(sensorContext);
+    FilePredicate filePredicate = analyzeAllFiles ? notBinaryFilePredicate : CombinedFilePredicate.any(LANGUAGE_FILE_PREDICATE, textFilePredicate);
```
jonas-wielage-sonarsource, line 82: CombinedFilePredicate can be replaced by `OR` on the predicate itself. Docker has an example for it

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextAndSecretsSensor.java
```diff
@@ -120,6 +101,10 @@ private static NotBinaryFilePredicate binaryFilePredicate(SensorContext sensorCo
     return new NotBinaryFilePredicate(sensorContext.config().getStringArray(TextAndSecretsSensor.EXCLUDED_FILE_SUFFIXES_KEY));
   }
 
+  private static TextFilePredicate textFilePredicate(SensorContext sensorContext) {
```
jonas-wielage-sonarsource, line 104: TextFilePredicate doesn't explain what is happening here.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextAndSecretsSensor.java
```diff
@@ -131,7 +116,7 @@ private static boolean analyzeAllFiles(SensorContext sensorContext) {
   /**
    * In SonarLint context we want to analyze all non-binary input files, even when they are not analyzed or assigned to a language.
    * To avoid analyzing all non-binary files to reduce time and memory consumption in a non SonarLint context only files assigned to a
-   * language are analyzed.
+   * language OR file with a text extension are analyzed.
```
jonas-wielage-sonarsource, line 119: text extension so `.txt` files?

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextFilePredicate.java
```diff
@@ -0,0 +1,57 @@
...
+    extensions = Arrays.stream(suffixes)
+      .map(String::trim)
+      .filter(value -> !value.isEmpty())
+      .collect(Collectors.toSet());
+  }
+
+  @Override
+  public boolean apply(InputFile inputFile) {
+    String filename = inputFile.filename();
+    String extension = extension(filename);
```
jonas-wielage-sonarsource, line 44: Shouldn't there already be a method to get the extension from the input File?

sonar-text-plugin/src/main/java/org/sonar/plugins/common/CombinedFilePredicate.java
```diff
@@ -0,0 +1,33 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.common;
+
+import java.util.Arrays;
+import org.sonar.api.batch.fs.FilePredicate;
+
+public final class CombinedFilePredicate {
```
jonas-wielage-sonarsource, line 25: I believe we don't need this class.

sonar-text-plugin/src/test/java/org/sonar/plugins/common/TextFilePredicateTest.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+package org.sonar.plugins.common;
+
+import java.nio.file.Path;
+import org.junit.jupiter.api.Test;
+import org.sonar.api.batch.fs.InputFile;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class TextFilePredicateTest {
```
jonas-wielage-sonarsource, line 28: This test could benefit from parametrized test

sonar-text-plugin/src/main/java/org/sonar/plugins/common/Analyzer.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+
+  private void analyzeFile(InputFile inputFile) {
+    try {
+      var inputFileContext = new InputFileContext(sensorContext, inputFile);
+      if (analyzeAllFilesMode) {
+        analyzeWithNotBinaryFileCheck(inputFileContext);
+      } else {
+        analyzeAllChecks(inputFileContext);
```
jonas-wielage-sonarsource, line 98: This is not how the original logic does it. 

The `notBinaryFilePredicate` is applied everytime in the `analyze` method of the Sensor.
The `notBinaryFilePredicate` is not applied `Text...Sensor.getInputFiles` when `analyzeAllFiles` is true.

In your example the usage is entirely dependent on the `analyzeAllFiles`.

its/plugin/src/test/java/org/sonarsource/text/FileDetectionTest.java
```diff
@@ -0,0 +1,79 @@
+package org.sonarsource.text;
+
+import com.sonar.orchestrator.build.SonarScanner;
+import javax.annotation.Nullable;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class FileDetectionTest extends TestBase {
```
jonas-wielage-sonarsource, line 9: Perfect candidates for parametrized tests

its/plugin/src/test/java/org/sonarsource/text/FileDetectionTest.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Test
+  public void shouldFindMultipleTextFileSpecifiedExtension() {
+    checkQuantityFileAnalyzedForConfiguration("mix-text-binary-files", false, "bash,txt", 2);
+  }
+
+  void checkQuantityFileAnalyzedForConfiguration(String subfolder, boolean analyzeAll, @Nullable String extensions, int quantityFileAnalyzed) {
+    // need to define a different project key for every check
+    String projectKey = PROJECT_KEY + counter++;
```
jonas-wielage-sonarsource, line 63: Not a big fan, but fine for me.

jonas-wielage-sonarsource (changes requested):

sonar-text-plugin/src/main/java/org/sonar/plugins/common/Analyzer.java
rudy-regazzoni-sonarsource: Mmhh indeed, the builder logic doesn't look really necessary in the end.
It made more sense in a previous version where I was iterating manually and calling the analyzer on each file, but now this logic was also moved in the `Analyzer` class. Now it's about instantiating a single object to then call a single method, doesn't really make sense.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextAndSecretsSensor.java
rudy-regazzoni-sonarsource: When I mention `text` file, it's not about `.txt` file, but file that contains human readable data, or non-binary data.
Whle we usually associate such file to the `.txt` extension, it could also be any other file extension/format that are designed to be human readable (`.json`, `.cpp`, ...)
This may be ambiguous, but I couldn't find any alternative/better term, and from my searches, it seems to be the way we express it.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextAndSecretsSensor.java
rudy-regazzoni-sonarsource: I wanted to have it similar to its cousin.
`NotBinaryFilePredicate` -> predicate to get file that are considered as not-binary file (blacklist approach)
`TextFilePredicate` -> predicate to get file that are considered as text file (whitelist approach)
I will add comments to both methods to explain the purpose of each.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextFilePredicate.java
rudy-regazzoni-sonarsource: Not sure what you mean here, `InputFile` provide a limited set of method, and nothing related to the extension itself, so we have to compute it ourselves.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/Analyzer.java
rudy-regazzoni-sonarsource: I believe I did it on purpose.

When `analyzeAllFiles` is `true`, the `FilePredicate` is `notBinaryFilePredicate`, so we are in blacklist mode -> we analyze all files that are not detected as a binary file
Here the check on file content makes sense, because with this blacklist mode, we may still find binary files that were not detected by their extension.

When `analyzeAllFiles` is `false`, the `FilePredicate` is `or(textFilePredicate, LANGUAGE_FILE_PREDICATE)`, so we are in whitelist mode and will only analyze files that either:
- has an associated language
- has a text file extension (xml, txt, json, ...) as defined by the property `sonar.text.included.file.suffixes`

The check on the file content doesn't make much sense here, we are in whitelist mode, as per the above restrictions we shouldn't encounter a binary file.

We could still perform the check, but in that case, we should adapt our message and display it for every case it appears.
Instead of this:
```
'java' was added to the binary file filter because the file 'MyClass.java' is a binary file.
To remove the previous warning you can add the '.java' extension to the 'sonar.text.excluded.file.suffixes' property.
```
(wrong property)

We could instead have something like this:
```
The file 'MyClass.java' contains binary data and will not be analyzed.
Please check this file and/or remove the extension from the 'sonar.text.included.file.suffixes' property.
```
WDYT?

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextAndSecretsSensor.java
jonas-wielage-sonarsource: Maybe simply `non-binary file`?
Or `file with an extension that from which we can conclude to contain readable text data are analyzed`.
Longer but not so ambiguous

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextFilePredicate.java
jonas-wielage-sonarsource: Let me have a look if I find something.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextAndSecretsSensor.java
rudy-regazzoni-sonarsource: non-binary may do the trick, but since we already have a `notBinaryFilePredicate`, the ambiguity may be worse :sweat_smile:
What about `plain-text file`? (`PlainTextFilePredicate`)

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextFilePredicate.java
jonas-wielage-sonarsource: Basically you don't need `TextFilePredicate`:

You could use something like this (but cleaned up, with more better collection handling.):
```
    List<FilePredicate> extensionPredicates = new ArrayList<>();
    for (String extension : extensions) {
      FilePredicate filePredicate = sensorContext.fileSystem().predicates().hasExtension(extension);
      extensionPredicates.add(filePredicate);
    }

    FilePredicate allExtensions = sensorContext.fileSystem().predicates().or(extensionPredicates);
```
This would replace your `textFilePredicate` in the Sensor.

sonar-text-plugin/src/main/java/org/sonar/plugins/common/Analyzer.java
jonas-wielage-sonarsource: Nice analysis of the context. I believe you're right.
I would like to keep this additional "safety net", and change the message as you have suggested.

its/plugin/src/test/java/org/sonarsource/text/FileDetectionTest.java
rudy-regazzoni-sonarsource: Yeah I don't like this either, but couldn't find a better solution :no_mouth:

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextAndSecretsSensor.java
jonas-wielage-sonarsource: Would be better.
Looking at the other comment below, I'm not sure we need `TextFilePredicate` afterall

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextFilePredicate.java
jonas-wielage-sonarsource: Good catch!

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextAndSecretsSensor.java
```diff
@@ -42,24 +37,20 @@
...
 
 public class TextAndSecretsSensor implements Sensor {
 
-  private static final Logger LOG = LoggerFactory.getLogger(TextAndSecretsSensor.class);
-
   public static final String EXCLUDED_FILE_SUFFIXES_KEY = "sonar.text.excluded.file.suffixes";
-
   private static final String ANALYZE_ALL_FILES_KEY = "sonar.text.analyzeAllFiles";
+  public static final String INCLUDED_FILE_SUFFIXES_KEY = "sonar.text.included.file.suffixes";
+  public static final String INCLUDED_FILE_SUFFIXES_DEFAULT_VALUE = "sh,bash,zsh,ksh,ps1,yaml,yml,properties,conf,xml,pem,env,config";
```
jonas-wielage-sonarsource, line 46:
```suggestion
  public static final String INCLUDED_FILE_SUFFIXES_DEFAULT_VALUE = "sh,bash,zsh,ksh,ps1,properties,conf,pem,env,config";
```
These properties are assigned to a language by default, so we don't need to include them

sonar-text-plugin/src/main/java/org/sonar/plugins/common/TextAndSecretsSensor.java
```diff
@@ -76,50 +67,56 @@ public void describe(SensorDescriptor descriptor) {
...
+    Analyzer.analyzeFiles(sensorContext, activeChecks, notBinaryFilePredicate, analyzeAllFiles, inputFiles);
     durationStatistics.log();
   }
 
-  private static NotBinaryFilePredicate binaryFilePredicate(SensorContext sensorContext) {
+  /**
+   * Blacklist approach: provide a predicate that exclude file that are considered as not-binary file.
+   * Example: for 'exe', 'txt' and 'unknown', it will return true for 'txt' and 'unknown'
+   * List of binary extension to exclude are provided by configuration key {@link TextAndSecretsSensor#EXCLUDED_FILE_SUFFIXES_KEY}
+   */
```
jonas-wielage-sonarsource, line 96: Nice descriptions overall in the PR

its/plugin/src/integrationTest/java/org/sonarsource/text/FileDetectionTest.java
```diff
@@ -0,0 +1,86 @@
...
+      assertThat(getMeasureAsInt(projectKey, "files")).isEqualTo(expectedNumberOfFileAnalyzed);
+    } else {
+      assertThat(getMeasureAsInt(projectKey, "files")).isNull();
+    }
+  }
+
+  @Parameterized.Parameters
+  public static Collection<Object> configurations() {
+    return Arrays.asList(new Object[][] {
+      {TEXT_FILES_ONLY_SUBFOLDER, true, null, 2},
```
jonas-wielage-sonarsource, line 76: I would like an additional tests, where we pass `""` and `" "` as `includedTextExtensions`

sonar-text-plugin/src/test/java/org/sonar/plugins/common/TextAndSecretsSensorTest.java
```diff
@@ -56,6 +56,7 @@
 class TextAndSecretsSensorTest {
 
   private static final String SENSITIVE_BIDI_CHARS = "\u0002\u0004";
+  private static final String INCLUDE_SUFFIX_KEY = "sonar.text.included.file.suffixes";
```
jonas-wielage-sonarsource, line 59: This key can be accessed by the sensor itself.

sonar-text-plugin/src/test/java/org/sonar/plugins/common/TextAndSecretsSensorTest.java
```diff
@@ -249,10 +250,60 @@ void shouldNotExecuteChecksOnBinaryFileNames() {
     assertThat(logTester.logs()).isEmpty();
   }
 
+  @Test
+  void shouldExecuteChecksOnIncludedTextFileNames() {
+    Check check = new ReportIssueAtLineOneCheck();
+    SensorContextTester context = sensorContext(check);
+    context.setSettings(new MapSettings().setProperty(INCLUDE_SUFFIX_KEY, "txt"));
```
jonas-wielage-sonarsource, line 257:
```suggestion
    context.setSettings(new MapSettings().setProperty(TextAndSecretsSensor.INCLUDED_FILE_SUFFIXES_KEY, "txt"));
```

sonar-text-plugin/src/test/resources/templates/rspecKeysToUpdate.txt
jonas-wielage-sonarsource, line 1: Remove this file...
At the current master, the file shouldn't be regenerated.

jonas-wielage-sonarsource (approved):
LGTM, please see my suggestions. Overall nice improvements!

sonar-text-plugin/src/test/resources/templates/rspecKeysToUpdate.txt
rudy-regazzoni-sonarsource: I thought I removed it already, sorry for this.

jonas-wielage-sonarsource:
Continuation will happen in SonarSource/sonar-text#14

jonas-wielage-sonarsource:
Continuation will happen in SonarSource/sonar-text#14

PR #302, opened 2023-10-04
SONARTEXT-69 Gradle fix ws scan

mstachniuk:


petertrr (approved):
LGTM, let's see if that helps

PR #303, opened 2023-10-04
(APPSEC-527) Filter out Firebase keys

gaetan-ferry-sonarsource:


gaetan-ferry-sonarsource:
The build task failed, I am not too sure why. That said, it failed late and only on the promote task so I guess that's OK.

sonar-text-plugin/src/main/resources/org/sonar/plugins/secrets/configuration/google-api.yaml
```diff
@@ -14,8 +14,12 @@ provider:
       metadata:
         name: Google API Keys
       detection:
+        pre:
+          reject:
+            content:
+              - "firebase"
         matching:
-          pattern: "\\b(AIza[\\w\\-]{35})\\b"
+          pattern: "\\b(AIza[0-9A-Za-z_\\-]{35})\\b"
```
loris-s-sonarsource, line 22: Hey, small FYI: \w is actually the same as what you replaced it with (even the _). Did you change it for clarity ?

gaetan-ferry-sonarsource: Yes. I have no issue replacing `[A-Za-z0-9_]` with `\w` but `[\\w\\-]` really looks weird to me. Moreover, I feel that `\w` as a meaning that does not fit the use here.
If you think we would be better reducing, fine by me.

loris-s-sonarsource: No totally fine by me also 👍

loris-s-sonarsource (approved):
LGTM !

jonas-wielage-sonarsource:
Will be continued on Continuation of SonarSource/sonar-text-enterprise#18

PR #304, opened 2023-10-04
SONARTEXT-69 Gradle add fastBuild task

mstachniuk:


build.gradle.kts
```diff
@@ -5,3 +5,13 @@ plugins {
...
 
 tasks.artifactoryPublish { skip = true }
+
+tasks.register("fastBuild") {
+  group = "Build"
+  description = "Runs fast build without integration tests"
+
+  dependsOn(tasks.build)
+  gradle.taskGraph.whenReady {
+    (gradle.taskGraph as org.gradle.execution.taskgraph.TaskExecutionGraphInternal).findTask("integrationTest")?.enabled = false
```
petertrr, line 15: Maybe we should exclude dependency between `check` and `integrationTest` and only run ITs explicitly?

petertrr (approved):
LGTM

PR #308, opened 2023-10-05
(APPSEC-1162) Reduce amount of FPs in test files with filename exclusions

gaetan-ferry-sonarsource:


sebastien-andrivet-sonarsource (approved):
Note: Checking some rules (such as Google OAuth) with SQ produces some Violating rule S6395: Unwrap this unnecessarily grouped subpattern.

